PR #90, opened 2016-11-30
Refactor

huangp:
This is in preparation for https://zanata.atlassian.net/browse/ZNTA-1527

- created a new module zanata-security to host shared code between zanata-module, zanata-war and zanata-test-war
- move the shared code to the new module 

## Reviewer tasks

The following is based on the
[Zanata Development Guidelines](https://github.com/zanata/zanata-server/wiki/Development-Guidelines).
to refresh your memory.

 - Code quality
   - If code is removed
 - Testing
   - [ ] New code has tests
   - [ ] Changed code has tests
   - [ ] All tests pass
 - Documentation
         [Considerations for new dependencies](https://github.com/zanata/zanata-server/wiki/Development-Guidelines#new-technologies-and-dependencies)?

----

---
This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-platform/90)


server/zanata-security/pom.xml
```diff
@@ -0,0 +1,35 @@
...
+  <packaging>takari-jar</packaging>
+  <url>http://maven.apache.org</url>
+
+  <!-- Should be the same as zanata-server/pom.xml -->
+  <scm>
+    <connection>scm:git:git://github.com/zanata/zanata.git</connection>
+    <developerConnection>scm:git:git@github.com:zanata/zanata.git</developerConnection>
+    <url>https://github.com/zanata/zanata</url>
+    <tag>HEAD</tag>
+  </scm>
```
seanf: This looks wrong. Is it needed?

server/zanata-security/pom.xml
```diff
@@ -0,0 +1,35 @@
...
+<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.zanata</groupId>
+    <artifactId>server</artifactId>
+    <version>4.1.0-SNAPSHOT</version>
+  </parent>
+  <artifactId>zanata-security</artifactId>
+  <name>zanata-security</name>
```
seanf: In terms of naming this `zanata-security`: right now this module just has the interfaces Identity, Authenticated and AuthenticatedLiteral. Is it likely to get most of the security code in future? It's not going to have three classes forever, so what are they going to have in common? I'm not sure zanata-authentication works either, given that our actual authentication code will probably be staying elsewhere.

I wonder if Identity should have a different name. That might inform the decision on the module name.

server/zanata-security/pom.xml
```diff
@@ -0,0 +1,35 @@
...
+
+  <!-- Should be the same as zanata-server/pom.xml -->
+  <scm>
+    <connection>scm:git:git://github.com/zanata/zanata.git</connection>
+    <developerConnection>scm:git:git@github.com:zanata/zanata.git</developerConnection>
+    <url>https://github.com/zanata/zanata</url>
+    <tag>HEAD</tag>
+  </scm>
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
```
seanf: This should be inherited from parent. Same goes for gwt-editor.

server/zanata-security/pom.xml
```diff
@@ -0,0 +1,35 @@
...
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.zanata</groupId>
+    <artifactId>server</artifactId>
+    <version>4.1.0-SNAPSHOT</version>
+  </parent>
+  <artifactId>zanata-security</artifactId>
+  <name>zanata-security</name>
+  <packaging>takari-jar</packaging>
+  <url>http://maven.apache.org</url>
```
seanf: The url (zanata.org) should just be inherited from parent. I just noticed that the same thing is in gwt-editor.

server/zanata-security/pom.xml
huangp: Most of our security code will still be in zanata-war. This is just a module to share some common interface and annotation. I am going to put more annotation into this module so that it can be shared. It will probably stay this way forever as long as we need to share code between modules. The zanata-security name came from one of the TODO tag but with a question mark at the end. I just took it and used it. Maybe security-common?

server/zanata-security/pom.xml
huangp: I am not sure. zanata-model and zanata-war both have these. But not in zanata-liuquibase. zanata-frontend has a different one
```
<!-- Connection URLs repeated as workaround for https://issues.apache.org/jira/browse/MPIR-234 -->
  <scm>
    <connection>scm:git:https://github.com/zanata/zanata-platform.git</connection>
    <developerConnection>scm:git:git@github.com:zanata/zanata-platform.git</developerConnection>
    <tag>HEAD</tag>
  </scm>
```
It's probably used by maven-release-plugin?

server/zanata-security/pom.xml
huangp: Generated with archetype:generate. Will remove them.

server/zanata-security/pom.xml
seanf: > It's probably used by maven-release-plugin?

Yes, that and/or the maven-site-plugin. Either we don't need this bit at all here, or we should copy the new one which mentions zanata-platform and MPIR-234.

server/zanata-security/pom.xml
seanf: Yes, security-common might work.

PR #175, opened 2017-02-02
test(ZNTA-1456): add unit tests for file adapters, instead of GUI tests

seanf:
https://zanata.atlassian.net/browse/ZNTA-1456

## Reviewer tasks

The following is based on the
[Zanata Development Guidelines](https://github.com/zanata/zanata-server/wiki/Development-Guidelines).
to refresh your memory.

 - Code quality
   - If code is removed
 - Testing
   - [ ] New code has tests
   - [ ] Changed code has tests
   - [ ] All tests pass
 - Documentation
         [Considerations for new dependencies](https://github.com/zanata/zanata-server/wiki/Development-Guidelines#new-technologies-and-dependencies)?

----

---
This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-platform/175)


seanf:
@djansen-redhat I think I've covered everything which was listed in the JIRA ticket, which should save about 10 minutes, but I think we can save another 22 minutes or so if you can convert FileTypeUploadTest and SubtitleDocumentTypeTest into SubtitleAdapterTest.

server/zanata-war/src/test/java/org/zanata/adapter/GettextAdapterTest.java
```diff
@@ -0,0 +1,64 @@
...
+        testFile = new File("src/test/resources/org/zanata/adapter/test-gettext.pot");
+        assert testFile.exists();
+    }
+
+    @Test
+    public void parsePOT() {
+        Resource resource =
+                adapter.parseDocumentFile(testFile.toURI(), LocaleId.EN,
+                        Optional.absent());
+//        System.out.println(DTOUtil.toXML(resource));
```
djansen-redhat: Should this be a log.debug or something?

seanf: We could, as long as the `DTOUtil.toXml` call is behind `if(log.isDebug())`. A bit cleaner, but I think it will make it slightly harder to use for debugging. This way, if you're working on the unit test you just uncomment one line. With `log.debug`, you'd have to tweak the logging config file to enable debug, and make sure the console appender threshold still allows debug. 

But it's up to you. Go ahead and make the change if you like.

seanf:
> This is great. Shall @spathare21 and I take it from here?

Yes @djansen-redhat, if you like. But I'm happy to support the changes I made if you have review comments to be addressed.

I was just hoping to get you to convert the various subtitle tests to adapter unit tests, since you're more familiar with that code and those tests, and maybe add the UI test you wanted for "will the multi-file upload form accept files with the right extensions?".

djansen-redhat:
@seanf sounds good to me

djansen-redhat:
@seanf subtitle f-test has been replaced. Your professional opinion appreciated.

server/zanata-war/src/test/java/org/zanata/adapter/SubtitleAdapterTest.java
```diff
@@ -0,0 +1,158 @@
...
+    /*
+     * Entries that are of identical content must be individual textflows
+     */
+    @Test
+    public void testSimilarEntriesAreIndividual() {
+        String testText = "Exactly the same text";
+        Resource resource = setupTestFile("test-srt-duplicated.srt");
+        assertThat(resource.getTextFlows()).hasSize(2);
+        assertThat(resource.getTextFlows().get(0).getId()).isNotEqualTo(
+                resource.getTextFlows().get(1).getId());
```
seanf: So you're not using a hash as the resId for de-duplication, like most of our other formats. What is it, positional, or does is it derived from the timestamp?

server/zanata-war/src/test/java/org/zanata/adapter/SubtitleAdapterTest.java
```diff
@@ -0,0 +1,158 @@
...
+                ImmutableList.of(testText));
+    }
+
+    /*
+     * Entries may contain multiple lines, and be preserved as such
+     */
+    @Test
+    public void testMultilineSubtitleFile() {
+        String testText = "Test subtitle 1\nTest subtitle 1 line 2";
+        String targetFile = "test-$-multiline.$";
```
seanf: I appreciate the lack of repetition, but it does make the test a bit hard to understand.

Any chance you could use a parameterised test[1], and just pass in the filename each time, and extract the extension from it?

[1] Please just make sure it doesn't run all the _other_ tests once for each data point. Or I suppose you could just have a for-loop over filenames, not fileTypes.

server/zanata-war/src/test/java/org/zanata/adapter/SubtitleAdapterTest.java
```diff
@@ -0,0 +1,158 @@
...
+        }
+    }
+
+    /*
+     * Entries may contain formatting - underline, bold, etc
+     */
+    @Test
+    public void testFormattedSubtitleFile() {
+        String testText = "<x1/>Some text<x2/> {u}and more{/u}";
+        String targetFile = "test-$-formatted.$";
```
seanf: Same here.

seanf:
@djansen-redhat Don't forget to eliminate the subtitle tests from FileTypeUploadTest too. I think we should convert the DTD test to a unit test too.

Did you have any review comments for me?

djansen-redhat:
@seanf thanks for the advice. I didn't have anything further to add on yours.
Yep, I've already started on removing the srt tests and replacing the dtd test.

server/zanata-war/src/test/java/org/zanata/adapter/SubtitleAdapterTest.java
```diff
@@ -0,0 +1,158 @@
...
+
+    private SubtitleAdapter adapter;
+    private final String[] fileTypes = new String[]{"srt", "vtt", "sbt", "sub"};
+
+    @Before
+    public void setup() {
+        adapter = new SubtitleAdapter();
+    }
+
+    private Resource setupTestFile(String filename) {
```
djansen-redhat: Done

server/zanata-war/src/test/java/org/zanata/adapter/SubtitleAdapterTest.java
djansen-redhat: Derived from the timestamp line, as far as I understand.

server/zanata-war/src/test/java/org/zanata/adapter/SubtitleAdapterTest.java
djansen-redhat: @seanf Like this? http://pastebin.test.redhat.com/452840

seanf: Yes, that looks pretty good, but instead of having the various test classes modify the static variable `adapter`, it would be a bit cleaner to give them all a common (static) superclass, which would hold `adapter`, `parseTestFile()` and `setup()`.

djansen-redhat: Done

server/zanata-war/src/test/java/org/zanata/adapter/SubtitleAdapterTest.java
djansen-redhat: Done

server/zanata-war/src/test/resources/org/zanata/adapter/test-gettext-comments.pot
```diff
@@ -0,0 +1,16 @@
+#  translator-comments
+#. comments-given-by-programmer
+#: reference
+#, flag
+msgid ""
```
djansen-redhat: Make the first entry Line One.
You should have a differing comment on the first and second entries and test them both.

server/zanata-war/src/test/java/org/zanata/adapter/GettextAdapterTest.java
```diff
@@ -0,0 +1,99 @@
...
+    @Test
+    public void testGettextWithComment() {
+        String testComments = "#  translator-comments\n" +
+                "#. comments-given-by-programmer\n" +
+                "#: reference\n" +
+                "#, flag";
+        Resource resource = parseTestFile("test-gettext-comments.pot");
+        assertThat(resource.getTextFlows()).hasSize(3);
+        assertThat(resource.getTextFlows().get(0).getContents()).isEqualTo(
+                ImmutableList.of(testComments));
```
djansen-redhat: This should not be testing contents, but testing something like:
PotEntryHeader potEntryHeader = (PotEntryHeader)textFlow.getExtensions().iterator().next();
assertThat(potEntryHeader.getFlags().get(0)).isEqualTo("flag");

The comment is also an extension, but cast it to SimpleComment.

server/zanata-war/src/test/java/org/zanata/adapter/GettextAdapterTest.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    @Test
+    public void testGettextPlurals(){
+        String testPlural = "Plural-Forms: nplurals=2; plural=n == 1 ? 0 : 1;\n";
+        Resource resource = parseTestFile("test-gettext-plurals.po");
+        assertThat(resource.getTextFlows()).hasSize(2);
+        assertThat(resource.getTextFlows().get(0).getContents()).isEqualTo(
+                ImmutableList.of(testPlural));
+        assertThat(resource.getTextFlows().get(1).getContents()).isEqualTo(
```
djansen-redhat: This should be:
size = 1
getContents() isEqualTo ImmutableList.of("%n file", "%n files"));

server/zanata-war/src/test/java/org/zanata/adapter/GettextAdapterTest.java
```diff
@@ -0,0 +1,99 @@
...
+                ImmutableList.of(testPlural));
+    }
+
+    @Test
+    public void testGettextFuzzy(){
+        String testFuzzy = "#, fuzzy";
+        Resource resource = parseTestFile("test-gettext-fuzzy.po");
+        assertThat(resource.getTextFlows()).hasSize(2);
+        assertThat(resource.getTextFlows().get(0).getContents()).isEqualTo(
+                ImmutableList.of(testFuzzy));
```
djansen-redhat: This is not correct, it should be the SimpleComment from the extensions I believe.

djansen-redhat:
@seanf tests are green. Unless there are any issues I think this is good to go.

Functional Test time reduced from approximately 1h 56m to 1h 21m.

server/zanata-war/src/test/java/org/zanata/adapter/SubtitleAdapterTest.java
```diff
@@ -33,126 +36,163 @@
...
+ * @author djansen <a href="mailto:djansen@redhat.com">djansen@redhat.com</a>
  * @// TODO: 6/02/17 test ids
  */
+@RunWith(Enclosed.class)
 public class SubtitleAdapterTest {
 
-    private SubtitleAdapter adapter;
-    private final String[] fileTypes = new String[]{"srt", "vtt", "sbt", "sub"};
+    public static abstract class SubtitleAdapterSuper {
+        private static SubtitleAdapter adapter;
```
seanf: - Package private should be enough; no need for public.
- The adapter field doesn't need to be static.
- Class name: how about AbstractSubtitleTest? It's not important though, especially if it becomes non-private.

server/functional-test/src/test/java/org/zanata/feature/document/MultiFileUploadTest.java
```diff
@@ -49,15 +54,15 @@
...
     public void before() {
         new BasicWorkFlow().goToHome().deleteCookiesAndRefresh();
         new LoginWorkFlow().signIn("admin", "admin");
         new ZanataRestCaller().createProjectAndVersion("multi-upload",
                 "multi-upload", "file");
-        documentStorageDirectory = CleanDocumentStorageRule
+        String documentStorageDirectory = CleanDocumentStorageRule
                 .getDocumentStoragePath().concat(File.separator)
                 .concat("documents").concat(File.separator);
         if (new File(documentStorageDirectory).exists()) {
```
seanf: How is this supposed to be cleaned up? Should we just delete it in this method?

server/functional-test/src/test/java/org/zanata/feature/document/UploadTest.java
```diff
@@ -50,15 +49,14 @@
...
     public void before() {
         new BasicWorkFlow().goToHome().deleteCookiesAndRefresh();
         new LoginWorkFlow().signIn("admin", "admin");
         new ZanataRestCaller().createProjectAndVersion("uploadtest",
                 "txt-upload", "file");
-        documentStorageDirectory = CleanDocumentStorageRule
+        String documentStorageDirectory = CleanDocumentStorageRule
                 .getDocumentStoragePath().concat(File.separator)
                 .concat("documents").concat(File.separator);
         if (new File(documentStorageDirectory).exists()) {
```
seanf: Again, it might be better just to clean up here, rather than risk a failed test.

server/zanata-war/src/test/java/org/zanata/adapter/DTDAdapterTest.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    @Test
+    public void parseBasicDTDFile() {
+        File testFile = new File("src/test/resources/org/zanata/adapter/basicdtd.dtd");
+        assert testFile.exists();
+        Resource resource =
+                adapter.parseDocumentFile(testFile.toURI(), LocaleId.EN,
+                        Optional.absent());
+        System.out.println(DTOUtil.toXML(resource));
```
seanf: This should be commented out. I think this is in one of the other tests too.

server/zanata-war/src/test/java/org/zanata/adapter/SubtitleAdapterTest.java
djansen-redhat: I think it does need to be static? The subclasses are static, and parse function is static.

server/functional-test/src/test/java/org/zanata/feature/document/MultiFileUploadTest.java
djansen-redhat: It's just a warning in case CleanDocumentStorageRule fails to do its job - which may be better in the actual rule but I'll look into that later (the check is in many tests)

server/functional-test/src/test/java/org/zanata/feature/document/UploadTest.java
djansen-redhat: ^see

server/zanata-war/src/test/java/org/zanata/adapter/DTDAdapterTest.java
djansen-redhat: Done (in both)

PR #192, opened 2017-02-13
test(adapter): Add JSONAdapter unit tests

djansen-redhat:
~~Fixed: The contentHash method allowed the content of two different key value pairs to be considered equal.~~
Update unit tests, add TODO for fixing id generation

## Reviewer tasks

The following is based on the
to refresh your memory.

 - Code quality
   - If code is removed
 - Testing
   - [ ] New code has tests
   - [ ] Changed code has tests
   - [ ] All tests pass
 - Documentation

----

---
This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-platform/192)

### DIFF
--- a/server/zanata-war/src/main/java/org/zanata/adapter/JsonAdapter.java
+++ b/server/zanata-war/src/main/java/org/zanata/adapter/JsonAdapter.java
@@ -21,8 +21,10 @@
 package org.zanata.adapter;
 
 import net.sf.okapi.common.IParameters;
+import net.sf.okapi.common.resource.TextUnit;
 import net.sf.okapi.filters.json.JSONFilter;
 import net.sf.okapi.filters.json.Parameters;
+import org.zanata.util.HashUtil;
 
 /**
  * Adapter to handle JavaScript Object Notation (JSON) documents.
@@ -45,5 +47,10 @@ public class JsonAdapter extends OkapiFilterAdapter {
     protected void updateParamsWithDefaults(IParameters params) {
         Parameters p = (Parameters) params;
         p.fromString("extractIsolatedStrings.b=true");
+    }
+
+    @Override
+    protected String getIdFor(TextUnit tu, String content, String subDocName) {
+        return HashUtil.generateHash(tu.getName().concat(content));
     }
 }
--- a/server/zanata-war/src/main/java/org/zanata/adapter/JsonAdapter.java
+++ b/server/zanata-war/src/main/java/org/zanata/adapter/JsonAdapter.java
@@ -21,10 +21,8 @@
 package org.zanata.adapter;
 
 import net.sf.okapi.common.IParameters;
-import net.sf.okapi.common.resource.TextUnit;
 import net.sf.okapi.filters.json.JSONFilter;
 import net.sf.okapi.filters.json.Parameters;
-import org.zanata.util.HashUtil;
 
 /**
  * Adapter to handle JavaScript Object Notation (JSON) documents.
@@ -49,8 +47,11 @@ public class JsonAdapter extends OkapiFilterAdapter {
         p.fromString("extractIsolatedStrings.b=true");
     }
 
+    /*
+     * TODO Get a global ID for the JSON fragment ZNTA-1731
     @Override
     protected String getIdFor(TextUnit tu, String content, String subDocName) {
         return HashUtil.generateHash(tu.getName().concat(content));
     }
+    */
 }
--- a/server/zanata-war/src/test/java/org/zanata/adapter/JsonAdapterTest.java
+++ b/server/zanata-war/src/test/java/org/zanata/adapter/JsonAdapterTest.java
@@ -23,43 +23,136 @@ package org.zanata.adapter;
 import static org.assertj.core.api.Assertions.assertThat;
 
 import java.io.File;
+import java.io.OutputStream;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 
+import com.google.common.base.Charsets;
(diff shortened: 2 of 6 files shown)


server/zanata-war/src/main/java/org/zanata/adapter/JsonAdapter.java
```diff
@@ -46,4 +48,9 @@ protected void updateParamsWithDefaults(IParameters params) {
         Parameters p = (Parameters) params;
         p.fromString("extractIsolatedStrings.b=true");
     }
+
+    @Override
+    protected String getIdFor(TextUnit tu, String content, String subDocName) {
+        return HashUtil.generateHash(tu.getName().concat(content));
```
djansen-redhat: I'm not sure that this was the right fix.  It:
- uses the key as part of the ID if it exists
- works afaik
- will not be compatible with currently uploaded translations

djansen-redhat:
@zanata-jenkins retest this please

server/zanata-war/src/main/java/org/zanata/adapter/JsonAdapter.java
seanf: Hmm.

Is that the local key, eg `title`, or some sort of global key, eg `test.test1.title` for a nested element? Global would be better.

I suppose any migration problems can be resolved by re-uploading the source and re-translating, although that could be a lot of work. Until then, a lot of translations would be missing when downloading.

As long as we still have the json file in the file system, we could implement a data migration which reparsed the json file and copied the translations to the new TextFlows. That's a lot of work though, considering we don't use json translations ourselves.

server/zanata-war/src/test/java/org/zanata/adapter/JsonAdapterTest.java
```diff
@@ -23,43 +23,136 @@
...
+import org.zanata.common.ContentState;
 import org.zanata.rest.dto.resource.Resource;
 
 import com.google.common.base.Optional;
-import com.google.common.collect.ImmutableList;
+import org.zanata.rest.dto.resource.TextFlowTarget;
 
 /**
  * @author Sean Flanigan <a href="mailto:sflaniga@redhat.com">sflaniga@redhat.com</a>
  */
```
seanf: You should add yourself as author.

server/zanata-war/src/test/java/org/zanata/adapter/JsonAdapterTest.java
```diff
@@ -23,43 +23,136 @@
...
+
+        TextFlowTarget firstTranslation = new TextFlowTarget();
+        firstTranslation.setContents("Foun’dé metalkcta");
+        firstTranslation.setState(ContentState.Approved);
+        translations.put("b76be7dbbd3aaf589a812e1a081d58ca", firstTranslation);
+
+        TextFlowTarget secondTranslation = new TextFlowTarget();
+        secondTranslation.setContents("Tba’dé metalkcta");
+        secondTranslation.setState(ContentState.Translated);
+        translations.put("da4cf86ebff968b8bde6ce056ee2e483", secondTranslation);
```
seanf: What are the magic hash codes?

seanf:
@djansen-redhat I wouldn't call this "chore" if it's changing the representation of parsed json files.

server/zanata-war/src/test/java/org/zanata/adapter/JsonAdapterTest.java
djansen-redhat: Multiple authors?

server/zanata-war/src/test/java/org/zanata/adapter/JsonAdapterTest.java
djansen-redhat: Those are the IDs for the translations, created through the generateHash method.

server/zanata-war/src/main/java/org/zanata/adapter/JsonAdapter.java
djansen-redhat: I thought maybe the global key would be overkill - but to be honest I'm not confident of this answer. I guess the question is - is it important enough to change it?
Also, could CopyTrans work for re-translating the document after update?

seanf: If you don't use a global key, won't some of your existing test cases (test-json-duplicateids.json) be merging similar strings into one text flow? I thought you wanted to avoid that.

CopyTrans should work if you tell it to ignore context, **in theory**.

server/zanata-war/src/test/java/org/zanata/adapter/JsonAdapterTest.java
seanf: That seems pretty fragile (not to mention opaque in terms of readability). It probably only makes sense to hard-code hashes like that in the unit test for the actual hash.

In this case, what if you parsed the json file and asked the relevant TextFlows for their IDs?

server/zanata-war/src/main/java/org/zanata/adapter/JsonAdapter.java
djansen-redhat: test-json-duplicateids.json won't merge as it requires similar key and content. The file test-json-duplicateentries.json however will (tested).
Looking at the data available in the tu at generating the ID, I don't even think I can provide the global id.

djansen-redhat:
@seanf last two commits are ready to be reviewed

seanf:
Reviewed 4 of 6 files at r1, 2 of 2 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/192)*

PR #193, opened 2017-02-13
Properties adapter tests

djansen-redhat:
JIRA issue URL: n/a

- test(adapter): Add properties tests
Test UTF-8 and Latin1 properties documents, the latter using generated files
- refactor(adapter): Relocate duplicated code
Move duplicated stream-in code to a util class
Use NonNull for parseDocument function rather than check and throw.

## Reviewer tasks

The following is based on the
to refresh your memory.

 - Code quality
   - If code is removed
 - Testing
   - [ ] New code has tests
   - [ ] Changed code has tests
   - [ ] All tests pass
 - Documentation

----

---
This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-platform/193)

### DIFF
--- a/server/zanata-war/src/main/java/org/zanata/adapter/FileFormatAdapter.java
+++ b/server/zanata-war/src/main/java/org/zanata/adapter/FileFormatAdapter.java
@@ -20,8 +20,12 @@
  */
 package org.zanata.adapter;
 
+import java.io.BufferedInputStream;
+import java.io.IOException;
 import java.io.OutputStream;
+import java.net.MalformedURLException;
 import java.net.URI;
+import java.net.URL;
 
 import javax.annotation.Nonnull;
 import javax.ws.rs.core.StreamingOutput;
@@ -138,5 +142,37 @@ public interface FileFormatAdapter {
         }
         return FilenameUtils.removeExtension(document
                 .getName()) + "." + transExt;
+    }
+
+    /**
+     * Read the data from the given file URI to a BufferedInputStream
+     *
+     * @param fileUri
+     * @return BufferedInputStream of file data
+     * @throws FileFormatAdapterException
+     * @throws IllegalArgumentException
+     */
+    default BufferedInputStream readStream(URI fileUri)
+            throws FileFormatAdapterException, IllegalArgumentException {
+        URL url = null;
+        try {
+            url = fileUri.toURL();
+            return new BufferedInputStream(url.openStream());
+        } catch (IllegalArgumentException e) {
+            throw new IllegalArgumentException(
+                    "Could not open the URI. The URI must be absolute: "
+                            + ((url == null) ? "URL is null" : url.toString()),
+                    e);
+        } catch (MalformedURLException e) {
+            throw new FileFormatAdapterException(
+                    "Could not open the URI. The URI may be malformed: "
+                            + ((url == null) ? "URL is null" : url.toString()),
+                    e);
+        } catch (IOException e) {
+            throw new FileFormatAdapterException(
+                    "Could not open the URL. The URL is OK but the input stream could not be opened.\n"
+                            + e.getMessage(),
+                    e);
+        }
     }
 }
--- a/server/zanata-war/src/main/java/org/zanata/adapter/FileFormatAdapter.java
+++ b/server/zanata-war/src/main/java/org/zanata/adapter/FileFormatAdapter.java
@@ -144,35 +144,4 @@ public interface FileFormatAdapter {
                 .getName()) + "." + transExt;
     }
 
-    /**
-     * Read the data from the given file URI to a BufferedInputStream
-     *
-     * @param fileUri
-     * @return BufferedInputStream of file data
-     * @throws FileFormatAdapterException
-     * @throws IllegalArgumentException
-     */
-    default BufferedInputStream readStream(URI fileUri)
-            throws FileFormatAdapterException, IllegalArgumentException {
-        URL url = null;
-        try {
-            url = fileUri.toURL();
-            return new BufferedInputStream(url.openStream());
-        } catch (IllegalArgumentException e) {
-            throw new IllegalArgumentException(
-                    "Could not open the URI. The URI must be absolute: "
-                            + ((url == null) ? "URL is null" : url.toString()),
-                    e);
-        } catch (MalformedURLException e) {
-            throw new FileFormatAdapterException(
-                    "Could not open the URI. The URI may be malformed: "
-                            + ((url == null) ? "URL is null" : url.toString()),
-                    e);
-        } catch (IOException e) {
-            throw new FileFormatAdapterException(
-                    "Could not open the URL. The URL is OK but the input stream could not be opened.\n"
-                            + e.getMessage(),
-                    e);
-        }
-    }
 }
--- a/server/zanata-war/src/test/java/org/zanata/adapter/PropertiesAbstractTest.java
+++ b/server/zanata-war/src/test/java/org/zanata/adapter/PropertiesAbstractTest.java
@@ -22,7 +22,7 @@ import static org.assertj.core.api.Assertions.assertThat;
 /**
  * @author Damian Jansen <a href="mailto:djansen@redhat.com">djansen@redhat.com</a>
  */
-class PropertiesAbstractTest {
+abstract class PropertiesAbstractTest {
 
     String resourcePath = "src/test/resources/org/zanata/adapter/";
     GenericPropertiesAdapter adapter;
@@ -45,16 +45,11 @@ class PropertiesAbstractTest {
     }
 
     File createTempFile(Charset charset) throws Exception {
-        File testFile;
-        try {
-            testFile = File.createTempFile("test-properties-latin1", "properties");
-        } catch (IOException ioException) {
-            throw new RuntimeException("Unable to create temporary file ");
-        }
+        File testFile = File.createTempFile("test-properties-temp", "properties");
         assertThat(testFile.exists());
         Map<String, String> entries = new HashMap<>();
         if (charset == StandardCharsets.ISO_8859_1) {
-            entries.put("line1", "Line One");
+            entries.put("line1", "^Line One");
         } else if (charset == StandardCharsets.UTF_8) {
             entries.put("line1", "¥Line One");
         }
--- a/server/zanata-war/src/test/java/org/zanata/adapter/PropertiesAbstractTest.java
+++ b/server/zanata-war/src/test/java/org/zanata/adapter/PropertiesAbstractTest.java
@@ -10,11 +10,10 @@ import org.zanata.rest.dto.resource.TranslationsResource;
 
 import java.io.File;
 import java.io.FileOutputStream;
-import java.io.IOException;
 import java.io.OutputStreamWriter;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
-import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.Map;
 
 import static org.assertj.core.api.Assertions.assertThat;
@@ -48,11 +47,15 @@ abstract class PropertiesAbstractTest {
         File testFile = File.createTempFile("test-properties-temp-" + charset, ".properties");
         System.out.println(testFile);
         assertThat(testFile.exists());
-        Map<String, String> entries = new HashMap<>();
+        Map<String, String> entries = new LinkedHashMap<>();
         if (charset == StandardCharsets.ISO_8859_1) {
             entries.put("line1", "ÀLine One");
+            entries.put("line2", "ÀLine Two");
+            entries.put("line3", "ÀLine Three");
         } else if (charset == StandardCharsets.UTF_8) {
             entries.put("line1", "¥Line One");
+            entries.put("line2", "¥Line Two");
+            entries.put("line3", "¥Line Three");
         }
         Properties resource = new Properties();
         for (Map.Entry<String, String> entry : entries.entrySet()) {
--- a/server/zanata-war/src/test/java/org/zanata/adapter/PropertiesAbstractTest.java
+++ b/server/zanata-war/src/test/java/org/zanata/adapter/PropertiesAbstractTest.java
@@ -45,11 +45,12 @@ abstract class PropertiesAbstractTest {
     }
 
     File createTempFile(Charset charset) throws Exception {
-        File testFile = File.createTempFile("test-properties-temp", "properties");
+        File testFile = File.createTempFile("test-properties-temp-" + charset, ".properties");
+        System.out.println(testFile);
         assertThat(testFile.exists());
         Map<String, String> entries = new HashMap<>();
         if (charset == StandardCharsets.ISO_8859_1) {
-            entries.put("line1", "^Line One");
+            entries.put("line1", "ÀLine One");
         } else if (charset == StandardCharsets.UTF_8) {
             entries.put("line1", "¥Line One");
         }
--- a/server/zanata-war/src/test/java/org/zanata/adapter/PropertiesAbstractTest.java
+++ b/server/zanata-war/src/test/java/org/zanata/adapter/PropertiesAbstractTest.java
@@ -45,7 +45,6 @@ abstract class PropertiesAbstractTest {
 
     File createTempFile(Charset charset) throws Exception {
         File testFile = File.createTempFile("test-properties-temp-" + charset, ".properties");
-        System.out.println(testFile);
         assertThat(testFile.exists());
         Map<String, String> entries = new LinkedHashMap<>();
         if (charset == StandardCharsets.ISO_8859_1) {
--- a/server/zanata-war/src/test/java/org/zanata/adapter/PropertiesAbstractTest.java
+++ b/server/zanata-war/src/test/java/org/zanata/adapter/PropertiesAbstractTest.java
@@ -1,0 +1,69 @@
+package org.zanata.adapter;
+
+import com.google.common.base.Optional;
+import org.fedorahosted.openprops.Properties;
+import org.zanata.common.ContentState;
+import org.zanata.common.LocaleId;
+import org.zanata.rest.dto.resource.Resource;
+import org.zanata.rest.dto.resource.TextFlowTarget;
+import org.zanata.rest.dto.resource.TranslationsResource;
+
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.OutputStreamWriter;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * @author Damian Jansen <a href="mailto:djansen@redhat.com">djansen@redhat.com</a>
+ */
+class PropertiesAbstractTest {
+
+    String resourcePath = "src/test/resources/org/zanata/adapter/";
+    GenericPropertiesAdapter adapter;
+
+    Resource parseTestFile(String fileName) {
+        File testFile = new File(resourcePath.concat(fileName));
+        assert testFile.exists();
+        return adapter.parseDocumentFile(testFile.toURI(), LocaleId.EN,
+                Optional.absent());
+    }
+
+    TranslationsResource addTranslation(TranslationsResource resource,
+                                        String id, String content, ContentState state) {
+        TextFlowTarget textFlowTarget = new TextFlowTarget();
+        textFlowTarget.setResId(id);
+        textFlowTarget.setContents(content);
+        textFlowTarget.setState(state);
+        resource.getTextFlowTargets().add(textFlowTarget);
+        return resource;
+    }
+
+    File createTempFile(Charset charset) throws Exception {
+        File testFile;
+        try {
+            testFile = File.createTempFile("test-properties-latin1", "properties");
+        } catch (IOException ioException) {
+            throw new RuntimeException("Unable to create temporary file ");
+        }
+        assertThat(testFile.exists());
+        Map<String, String> entries = new HashMap<>();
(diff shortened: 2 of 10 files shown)


server/zanata-war/src/main/java/org/zanata/adapter/FileFormatAdapter.java
```diff
@@ -139,4 +143,36 @@ default String generateTranslationFilename(@Nonnull HDocument document, @Nonnull
...
+
+    /**
+     * Read the data from the given file URI to a BufferedInputStream
+     *
+     * @param fileUri
+     * @return BufferedInputStream of file data
+     * @throws FileFormatAdapterException
+     * @throws IllegalArgumentException
+     */
+    default BufferedInputStream readStream(URI fileUri)
```
seanf: This method doesn't belong in this interface. We should put it in a package-private utility function, because we shouldn't even be passing around file URIs, really.

server/zanata-war/src/test/java/org/zanata/adapter/PropertiesAbstractTest.java
```diff
@@ -0,0 +1,69 @@
...
+import java.nio.charset.StandardCharsets;
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * @author Damian Jansen <a href="mailto:djansen@redhat.com">djansen@redhat.com</a>
+ */
+class PropertiesAbstractTest {
```
seanf: Shouldn't this be an `abstract class`?

server/zanata-war/src/test/java/org/zanata/adapter/PropertiesAbstractTest.java
```diff
@@ -0,0 +1,69 @@
...
+        resource.getTextFlowTargets().add(textFlowTarget);
+        return resource;
+    }
+
+    File createTempFile(Charset charset) throws Exception {
+        File testFile;
+        try {
+            testFile = File.createTempFile("test-properties-latin1", "properties");
+        } catch (IOException ioException) {
+            throw new RuntimeException("Unable to create temporary file ");
```
seanf: If this is test code, why bother catching the IOException? You could just let it propagate.

server/zanata-war/src/test/java/org/zanata/adapter/PropertiesAbstractTest.java
```diff
@@ -0,0 +1,69 @@
...
+        }
+        assertThat(testFile.exists());
+        Map<String, String> entries = new HashMap<>();
+        if (charset == StandardCharsets.ISO_8859_1) {
+            entries.put("line1", "Line One");
```
seanf: Don't you want to put some Latin 1 characters in the file?

server/zanata-war/src/test/java/org/zanata/adapter/PropertiesAbstractTest.java
```diff
@@ -0,0 +1,69 @@
...
+        textFlowTarget.setContents(content);
+        textFlowTarget.setState(state);
+        resource.getTextFlowTargets().add(textFlowTarget);
+        return resource;
+    }
+
+    File createTempFile(Charset charset) throws Exception {
+        File testFile;
+        try {
+            testFile = File.createTempFile("test-properties-latin1", "properties");
```
seanf: The filename "test-properties-latin1" could be misleading if `charset` is not Latin 1.

server/zanata-war/src/test/java/org/zanata/adapter/PropertiesLatinOneAdapterTest.java
```diff
@@ -22,44 +22,104 @@
...
+        Resource resource = parseTestFile("test-properties-latin1.properties");
         assertThat(resource.getTextFlows()).hasSize(3);
         assertThat(resource.getTextFlows().get(0).getId()).isEqualTo(
                 "line1");
         assertThat(resource.getTextFlows().get(0).getContents()).isEqualTo(
                 ImmutableList.of("Line One"));
     }
 
+    /*
+     * Properties files change path, not name
```
seanf: Huh? What's this testing?

server/zanata-war/src/test/java/org/zanata/adapter/PropertiesLatinOneAdapterTest.java
```diff
@@ -22,44 +22,104 @@
...
+    @Test
+    public void testUTFOnLatin1encoding() throws Exception {
+        File latin1EncodedFile = createTempFile(StandardCharsets.UTF_8);
+        Resource resource =
+                adapter.parseDocumentFile(latin1EncodedFile.toURI(), LocaleId.EN,
+                        Optional.absent());
+        assertThat(resource.getTextFlows().get(0).getId()).isEqualTo(
+                "line1");
+        assertThat(resource.getTextFlows().get(0).getContents())
+                .containsExactly("Â¥Line One");
```
seanf: I don't think it's a good idea to test that a file gets corrupted with a particular corruption. Better to test that we can read and write the character "¾" correctly via Latin 1, for instance.

server/zanata-war/src/test/java/org/zanata/adapter/PropertiesUTF8AdapterTest.java
```diff
@@ -23,43 +23,56 @@
...
+
+        adapter.writeTranslatedFile(outputStream,
+                originalFile.toURI(),
+                resource,
+                tResource,
+                "ru",
+                Optional.absent());
+
+        assertThat(outputStream.toString()).isEqualTo(
+                "line1=¥Foun’dé metalkcta\nline2=¥Tba’dé metalkcta\nline3=\n");
```
seanf: This would be easier to read if you split the string after each `\n` (even if we can't control the line wrapping very well).

server/zanata-war/src/main/java/org/zanata/adapter/FileFormatAdapter.java
djansen-redhat: Truth. Fixed.

server/zanata-war/src/test/java/org/zanata/adapter/PropertiesAbstractTest.java
djansen-redhat: Yes...

server/zanata-war/src/test/java/org/zanata/adapter/PropertiesAbstractTest.java
djansen-redhat: Oops.

server/zanata-war/src/test/java/org/zanata/adapter/PropertiesAbstractTest.java
djansen-redhat: I suppose the error would be obvious, but no you're right.

server/zanata-war/src/test/java/org/zanata/adapter/PropertiesLatinOneAdapterTest.java
djansen-redhat: The generation of file name?

server/zanata-war/src/test/java/org/zanata/adapter/PropertiesAbstractTest.java
```diff
@@ -45,16 +45,11 @@ TranslationsResource addTranslation(TranslationsResource resource,
...
-            testFile = File.createTempFile("test-properties-latin1", "properties");
-        } catch (IOException ioException) {
-            throw new RuntimeException("Unable to create temporary file ");
-        }
+        File testFile = File.createTempFile("test-properties-temp", "properties");
         assertThat(testFile.exists());
         Map<String, String> entries = new HashMap<>();
         if (charset == StandardCharsets.ISO_8859_1) {
-            entries.put("line1", "Line One");
+            entries.put("line1", "^Line One");
```
seanf: I don't think `circumflex accent` is the ideal choice, because it seems to have the same representation in UTF-8. Try `À` which is `0xC0` in Latin-1. `0xC0` is illegal in UTF-8.

djansen-redhat:
*[server/zanata-war/src/test/java/org/zanata/adapter/PropertiesAbstractTest.java, line 57 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/193#-KcrPZF3TWxSjFfGzDWW-r1-57:-KcuxDHvlMDcNK-oep_M:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/223fdaa19e74d0853d688f746a7ed86de29a8f66/server/zanata-war/src/test/java/org/zanata/adapter/PropertiesAbstractTest.java#L57)):*

Don't you want to put some Latin 1 characters in the file?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/193)*

djansen-redhat:
*[server/zanata-war/src/test/java/org/zanata/adapter/PropertiesAbstractTest.java, line 52 at r2](https://reviewable.io:443/reviews/zanata/zanata-platform/193#-KcrPZF3TWxSjFfGzDWW-r2-52:-KcuxElpvzDHBNl9cYz8:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/d05edff437be384a9a9bfd4a15fbf3d50c826695/server/zanata-war/src/test/java/org/zanata/adapter/PropertiesAbstractTest.java#L52)):*

I don't think `circumflex accent` is the ideal choice, because it seems to have the same representation in UTF-8. Try `À` which is `0xC0` in Latin-1. `0xC0` is illegal in UTF-8.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/193)*

djansen-redhat:
*[server/zanata-war/src/test/java/org/zanata/adapter/PropertiesLatinOneAdapterTest.java, line 122 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/193#-KcrPZF3TWxSjFfGzDWX-r1-122:-KcuxKcwRscIEi7fiNN7:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/223fdaa19e74d0853d688f746a7ed86de29a8f66/server/zanata-war/src/test/java/org/zanata/adapter/PropertiesLatinOneAdapterTest.java#L122)):*

I don't think it's a good idea to test that a file gets corrupted with a particular corruption. Better to test that we can read and write the character "¾" correctly via Latin 1, for instance.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/193)*

djansen-redhat:
*[server/zanata-war/src/test/java/org/zanata/adapter/PropertiesUTF8AdapterTest.java, line 75 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/193#-KcrPZF3TWxSjFfGzDWY-r1-75:-KcuxNqDJ9behIUnzng8:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/223fdaa19e74d0853d688f746a7ed86de29a8f66/server/zanata-war/src/test/java/org/zanata/adapter/PropertiesUTF8AdapterTest.java#L75)):*

This would be easier to read if you split the string after each `\n` (even if we can't control the line wrapping very well).
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/193)*

seanf:
Reviewed 2 of 8 files at r1, 5 of 8 files at r2, 3 of 3 files at r3.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[server/zanata-war/src/test/java/org/zanata/adapter/PropertiesLatinOneAdapterTest.java, line 57 at r3](https://reviewable.io:443/reviews/zanata/zanata-platform/193#-Kd-Y8gJlwqdanaT-Etz:-Kd-Y8gJlwqdanaT-Eu-:b-jhj71o) ([raw file](https://github.com/zanata/zanata-platform/blob/d98d6656e2788d9d2b0cc3c4d0fb2df9f2bc2972/server/zanata-war/src/test/java/org/zanata/adapter/PropertiesLatinOneAdapterTest.java#L57)):*
> ```Java
>     @Test
>     public void parseLatinOneProperties() {
>         Resource resource = parseTestFile("test-properties-latin1.properties");
> ```

I'm a bit concerned that the Latin-1 encoded file test-properties-latin1.properties may accidentally change its format at some point. Now that you have tests which generate Latin-1 files on the fly, perhaps we could delete this test.

---

*[server/zanata-war/src/test/java/org/zanata/adapter/PropertiesLatinOneAdapterTest.java, line 78 at r3](https://reviewable.io:443/reviews/zanata/zanata-platform/193#-Kd-X5msrT3rAPTfh6hm:-Kd-X5msrT3rAPTfh6hn:bcm884r) ([raw file](https://github.com/zanata/zanata-platform/blob/d98d6656e2788d9d2b0cc3c4d0fb2df9f2bc2972/server/zanata-war/src/test/java/org/zanata/adapter/PropertiesLatinOneAdapterTest.java#L78)):*
> ```Java
> 
>         assertThat(adapter.generateTranslationFilename(document, "fr"))
>                 .isEqualTo("basicprop.properties");
> ```

I'm still not sure I understand the purpose of this test. Shouldn't the name of the `fr` translation of basicprop.properties be basicprop_fr.properties? That's the usual rule for translated properties files. Or is all this just testing that the filename is unchanged? If so, it seems like a long-winded way of testing that a = a.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/193)*

djansen-redhat:
*[server/zanata-war/src/test/java/org/zanata/adapter/PropertiesLatinOneAdapterTest.java, line 78 at r3](https://reviewable.io:443/reviews/zanata/zanata-platform/193#-Kd-X5msrT3rAPTfh6hm:-Kd3WBPz65uhgPpl7bfZ:b-ak7z8i) ([raw file](https://github.com/zanata/zanata-platform/blob/d98d6656e2788d9d2b0cc3c4d0fb2df9f2bc2972/server/zanata-war/src/test/java/org/zanata/adapter/PropertiesLatinOneAdapterTest.java#L78)):*

I'm still not sure I understand the purpose of this test. Shouldn't the name of the `fr` translation of basicprop.properties be basicprop_fr.properties? That's the usual rule for translated properties files. Or is all this just testing that the filename is unchanged? If so, it seems like a long-winded way of testing that a = a.
</blockquote></details>

I don't see a problem with tests that say "if I put this a into b, I should get a back out again".
But are you suggesting the implementation is actually wrong?  I get the same filename as the original when I download properties files via WebUI.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/193)*

djansen-redhat:
*[server/zanata-war/src/test/java/org/zanata/adapter/PropertiesLatinOneAdapterTest.java, line 57 at r3](https://reviewable.io:443/reviews/zanata/zanata-platform/193#-Kd-Y8gJlwqdanaT-Etz:-Kd3VGfZdYKFUD6e3kyC:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/d98d6656e2788d9d2b0cc3c4d0fb2df9f2bc2972/server/zanata-war/src/test/java/org/zanata/adapter/PropertiesLatinOneAdapterTest.java#L57)):*

I'm a bit concerned that the Latin-1 encoded file test-properties-latin1.properties may accidentally change its format at some point. Now that you have tests which generate Latin-1 files on the fly, perhaps we could delete this test.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/193)*

seanf:
Reviewed 3 of 3 files at r4.

---

*[server/zanata-war/src/test/java/org/zanata/adapter/PropertiesLatinOneAdapterTest.java, line 78 at r3](https://reviewable.io:443/reviews/zanata/zanata-platform/193#-Kd-X5msrT3rAPTfh6hm:-Kd3YowALyDQ3zGHT0X4:b9qn4xh) ([raw file](https://github.com/zanata/zanata-platform/blob/d98d6656e2788d9d2b0cc3c4d0fb2df9f2bc2972/server/zanata-war/src/test/java/org/zanata/adapter/PropertiesLatinOneAdapterTest.java#L78)):*
<details><summary><i>Previously, djansen-redhat (Damian Jansen) wrote…</i></summary><blockquote>

I don't see a problem with tests that say "if I put this a into b, I should get a back out again".
But are you suggesting the implementation is actually wrong?  I get the same filename as the original when I download properties files via WebUI.
</blockquote></details>

I think it might be. 

I'm not sure what the intended use case for downloading a .properties file via the UI is, except for completeness. If you have software to translate, wouldn't you use the cli tools?  

But I can't think of any good reason to name the translation file the same as the source .properties file, just that it would be the easiest option. In a Java program, if you want to pick up the French translations for `something.properties`, it's only going to work if the filename is `something_fr.properties`.

It may not have been planned this way. It could be just that the other formats were implemented first, and in general I don't think they need translations to be renamed.

Do we have a similar test for the other file types?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/193)*

djansen-redhat:
*[server/zanata-war/src/test/java/org/zanata/adapter/PropertiesLatinOneAdapterTest.java, line 78 at r3](https://reviewable.io:443/reviews/zanata/zanata-platform/193#-Kd-X5msrT3rAPTfh6hm:-Kd3zX_vnAFZo27F9fXi:blmfzed) ([raw file](https://github.com/zanata/zanata-platform/blob/d98d6656e2788d9d2b0cc3c4d0fb2df9f2bc2972/server/zanata-war/src/test/java/org/zanata/adapter/PropertiesLatinOneAdapterTest.java#L78)):*

I think it might be. 

I'm not sure what the intended use case for downloading a .properties file via the UI is, except for completeness. If you have software to translate, wouldn't you use the cli tools?  

But I can't think of any good reason to name the translation file the same as the source .properties file, just that it would be the easiest option. In a Java program, if you want to pick up the French translations for `something.properties`, it's only going to work if the filename is `something_fr.properties`.

It may not have been planned this way. It could be just that the other formats were implemented first, and in general I don't think they need translations to be renamed.

Do we have a similar test for the other file types?
</blockquote></details>

So I had a bit of a look into it - the file naming is done directly by the pull strategy classes _for the client_, in that it expects a document name and adds the locale mapping and extension.
The WebUI on the other hand does not, save for my implementation of TSAdapter which handles it itself.
And _then_, if you're using a File type project to push your .properties files to, pull defaults to a locale directory structure like gettext.
A similar behaviour if you use WebUI to download all as zip, File or Properties project.
This is so messy!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/193)*

seanf:
Reviewed 2 of 3 files at r5.
Review status: 9 of 10 files reviewed at latest revision, 1 unresolved discussion.

---

*[server/zanata-war/src/main/java/org/zanata/adapter/GenericPropertiesAdapter.java, line 152 at r5](https://reviewable.io:443/reviews/zanata/zanata-platform/193#-Kd4G_3JdQgHN7nCH0_H:-Kd4G_3KjPUsHYyX4h6d:b-sq1jc0) ([raw file](https://github.com/zanata/zanata-platform/blob/8234f2bc9d51b93755ebdede20fd112cbed39a71/server/zanata-war/src/main/java/org/zanata/adapter/GenericPropertiesAdapter.java#L152)):*
> ```Java
>                     localeId.toJavaName() + ".properties";
>         }
>         return FilenameUtils.removeExtension(document.getName()) + "_"
> ```

Given the code above, I have to ask what happens if the srcExt is empty? (Not that I've ever seen a Properties file without the .properties extension.)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/193)*

djansen-redhat:
@zanata-jenkins retest this please

djansen-redhat:
*[server/zanata-war/src/main/java/org/zanata/adapter/GenericPropertiesAdapter.java, line 152 at r5](https://reviewable.io:443/reviews/zanata/zanata-platform/193#-Kd4G_3JdQgHN7nCH0_H:-Kd58SZztAK19RvB4I-n:b-dozrbm) ([raw file](https://github.com/zanata/zanata-platform/blob/8234f2bc9d51b93755ebdede20fd112cbed39a71/server/zanata-war/src/main/java/org/zanata/adapter/GenericPropertiesAdapter.java#L152)):*

Given the code above, I have to ask what happens if the srcExt is empty? (Not that I've ever seen a Properties file without the .properties extension.)
</blockquote></details>

If it's empty, it just adds it on. Question is, what if it's empty, and the name ends in '.'? You get a filename of filename._fr.properties - valid?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/193)*

seanf:
:lgtm:

---

Reviewed 1 of 3 files at r5.

---

*[server/zanata-war/src/main/java/org/zanata/adapter/GenericPropertiesAdapter.java, line 152 at r5](https://reviewable.io:443/reviews/zanata/zanata-platform/193#-Kd4G_3JdQgHN7nCH0_H:-Kd9-pWPjeZ0ZwtGyNMo:b-jlntln) ([raw file](https://github.com/zanata/zanata-platform/blob/8234f2bc9d51b93755ebdede20fd112cbed39a71/server/zanata-war/src/main/java/org/zanata/adapter/GenericPropertiesAdapter.java#L152)):*
<details><summary><i>Previously, djansen-redhat (Damian Jansen) wrote…</i></summary><blockquote>

If it's empty, it just adds it on. Question is, what if it's empty, and the name ends in '.'? You get a filename of filename._fr.properties - valid?
</blockquote></details>

If the user has code that doesn't mind a weird properties filename ending in `.`, I don't see why they couldn't handle another weird `filename._fr.properties` just as well.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/193#-:-Kd90Lzp12THfd1rpGAm:bnfp4nl)*

PR #255, opened 2017-03-31
ZNTA-1760 - modify constraint and fix functional test

huangp:
This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-platform/255)


seanf:
@huangp What changes did you make to FK_HProject_LocaleMember_HLocale?

seanf:
@huangp Oh, I see you just added `onDelete="CASCADE"` - so what does that do exactly, when applied to a join table?

huangp:
@seanf it means it someone delete a HLocale, it will delete the rows in HProject_LocaleMember. I thought we don't support language delete (see the comment in previous changelog) but I was wrong. We do support it as long as there is no translation. So I thought it's better to add the cascade rule there too. Same as HPerson.

djansen-redhat:
✅ looks ok to me

PR #362, opened 2017-06-02
ZNTA-2022 - change HTextFlowTarget index strategy

huangp:
- exclude HTextFlowTargets from obsolete project or version
- index project id instead of slug (so it's immutable)
- index version id instead of slug (so it's immutable)
- reindex when project or version become obsolete

JIRA issue URL: https://zanata.atlassian.net/browse/ZNTA-2022

note to QA:
- need to perform re-index HTextFlowTarget in administration-> manage search
- if a project or version is deleted (status become obsolete in database), there should be a re-index log message in console. It should be transparent to user
- TM merge should still work as before. Need to make sure when rejecting TM from different project, it does what it says. e.g. when you choose reject TM from different project, you should not get TM from different project

---
This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-platform/362)

### DIFF
--- a/server/zanata-model/src/main/java/org/zanata/hibernate/search/ContainingWorkspaceBridge.java
+++ b/server/zanata-model/src/main/java/org/zanata/hibernate/search/ContainingWorkspaceBridge.java
@@ -40,9 +40,10 @@ public class ContainingWorkspaceBridge extends AbstractFieldBridge {
         HProjectIteration iteration = doc.getProjectIteration();
         HProject project = iteration.getProject();
 
-        addStringField(IndexFieldLabels.PROJECT_FIELD, project.getSlug(),
+        addStringField(IndexFieldLabels.PROJECT_FIELD, project.getId().toString(),
                 luceneDocument, luceneOptions);
-        addStringField(IndexFieldLabels.PROJECT_VERSION_FIELD, iteration.getSlug(),
+        addStringField(IndexFieldLabels.PROJECT_VERSION_FIELD,
+                iteration.getId().toString(),
                 luceneDocument, luceneOptions);
         addStringField(IndexFieldLabels.DOCUMENT_ID_FIELD, doc.getDocId(),
                 luceneDocument, luceneOptions);
--- a/server/zanata-model/src/main/java/org/zanata/hibernate/search/ContainingWorkspaceBridge.java
+++ b/server/zanata-model/src/main/java/org/zanata/hibernate/search/ContainingWorkspaceBridge.java
@@ -14,8 +14,8 @@ import org.zanata.model.HTextFlow;
  *
  * The provided field name is not used
  * <ul>
- * <li>Project slug is indexed as 'project'</li>
- * <li>Iteration slug is indexed as 'projectVersion'</li>
+ * <li>Project id is indexed as 'project'</li>
+ * <li>Iteration id is indexed as 'projectVersion'</li>
  * <li>Document full path + name is indexed as 'documentId'</li>
  * </ul>
  *
@@ -40,9 +40,9 @@ public class ContainingWorkspaceBridge extends AbstractFieldBridge {
         HProjectIteration iteration = doc.getProjectIteration();
         HProject project = iteration.getProject();
 
-        addStringField(IndexFieldLabels.PROJECT_FIELD, project.getId().toString(),
+        addStringField(IndexFieldLabels.PROJECT_ID_FIELD, project.getId().toString(),
                 luceneDocument, luceneOptions);
-        addStringField(IndexFieldLabels.PROJECT_VERSION_FIELD,
+        addStringField(IndexFieldLabels.PROJECT_VERSION_ID_FIELD,
                 iteration.getId().toString(),
                 luceneDocument, luceneOptions);
         addStringField(IndexFieldLabels.DOCUMENT_ID_FIELD, doc.getDocId(),
--- a/server/zanata-model/src/main/java/org/zanata/hibernate/search/EntityStatusIndexingInterceptor.java
+++ b/server/zanata-model/src/main/java/org/zanata/hibernate/search/EntityStatusIndexingInterceptor.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright 2017, Red Hat, Inc. and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.zanata.hibernate.search;
+
+import org.hibernate.search.indexes.interceptor.EntityIndexingInterceptor;
+import org.hibernate.search.indexes.interceptor.IndexingOverride;
+import org.zanata.common.EntityStatus;
+import org.zanata.model.HProjectIteration;
+import org.zanata.model.HTextFlowTarget;
+
+/**
+ * @author Patrick Huang <a href="mailto:pahuang@redhat.com">pahuang@redhat.com</a>
+ */
+public class EntityStatusIndexingInterceptor implements
+        EntityIndexingInterceptor<HTextFlowTarget> {
+    @Override
+    public IndexingOverride onAdd(HTextFlowTarget hTextFlowTarget) {
+        if (hTextFlowTarget.getTextFlow().isObsolete()) {
+            return IndexingOverride.SKIP;
+        }
+        if (hTextFlowTarget.getTextFlow().getDocument().isObsolete()) {
+            return IndexingOverride.SKIP;
+        }
+        HProjectIteration projectIteration =
+                hTextFlowTarget.getTextFlow().getDocument()
+                        .getProjectIteration();
+        if (projectIteration.getStatus() == EntityStatus.OBSOLETE) {
+            return IndexingOverride.SKIP;
+        }
+        if (projectIteration.getProject().getStatus() == EntityStatus.OBSOLETE) {
+            return IndexingOverride.SKIP;
+        }
+        return IndexingOverride.APPLY_DEFAULT;
+    }
+
+    @Override
+    public IndexingOverride onUpdate(HTextFlowTarget hTextFlowTarget) {
+        if (hTextFlowTarget.getTextFlow().isObsolete()) {
+            return IndexingOverride.REMOVE;
+        }
+        if (hTextFlowTarget.getTextFlow().getDocument().isObsolete()) {
+            return IndexingOverride.REMOVE;
+        }
+        HProjectIteration projectIteration =
+                hTextFlowTarget.getTextFlow().getDocument()
+                        .getProjectIteration();
+        if (projectIteration.getStatus() == EntityStatus.OBSOLETE) {
+            return IndexingOverride.REMOVE;
+        }
+        if (projectIteration.getProject().getStatus() == EntityStatus.OBSOLETE) {
+            return IndexingOverride.REMOVE;
+        }
+        return IndexingOverride.UPDATE;
+    }
+
+    @Override
+    public IndexingOverride onDelete(HTextFlowTarget hTextFlowTarget) {
+        return IndexingOverride.APPLY_DEFAULT;
+    }
+
+    @Override
+    public IndexingOverride onCollectionUpdate(
+            HTextFlowTarget hTextFlowTarget) {
+        return onUpdate(hTextFlowTarget);
+    }
+}
--- a/server/zanata-model/src/main/java/org/zanata/hibernate/search/IndexFieldLabels.java
+++ b/server/zanata-model/src/main/java/org/zanata/hibernate/search/IndexFieldLabels.java
@@ -7,7 +7,7 @@ package org.zanata.hibernate.search;
  * @author David Mason, damason@redhat.com
  */
 public interface IndexFieldLabels {
-    public static final String PROJECT_FIELD = "project";
+    public static final String PROJECT_ID_FIELD = "project";
     public static final String ENTITY_STATUS = "status";
 
     /**
@@ -35,5 +35,5 @@ public interface IndexFieldLabels {
 
     public static final String GLOSSARY_QUALIFIED_NAME = "glossaryEntry.glossary.qualifiedName";
 
-    String PROJECT_VERSION_FIELD = "projectVersion";
+    String PROJECT_VERSION_ID_FIELD = "projectVersion";
 }
--- a/server/zanata-model/src/main/java/org/zanata/hibernate/search/IndexFieldLabels.java
+++ b/server/zanata-model/src/main/java/org/zanata/hibernate/search/IndexFieldLabels.java
@@ -7,7 +7,7 @@ package org.zanata.hibernate.search;
  * @author David Mason, damason@redhat.com
  */
 public interface IndexFieldLabels {
-    public static final String PROJECT_ID_FIELD = "project";
+    public static final String PROJECT_ID_FIELD = "projectId";
     public static final String ENTITY_STATUS = "status";
 
     /**
@@ -35,5 +35,5 @@ public interface IndexFieldLabels {
 
     public static final String GLOSSARY_QUALIFIED_NAME = "glossaryEntry.glossary.qualifiedName";
 
-    String PROJECT_VERSION_ID_FIELD = "projectVersion";
+    String PROJECT_VERSION_ID_FIELD = "projectVersionId";
 }
--- a/server/zanata-model/src/main/java/org/zanata/hibernate/search/TextFlowTargetEntityIndexingInterceptor.java
+++ b/server/zanata-model/src/main/java/org/zanata/hibernate/search/TextFlowTargetEntityIndexingInterceptor.java
@@ -29,7 +29,7 @@ import org.zanata.model.HTextFlowTarget;
 /**
  * @author Patrick Huang <a href="mailto:pahuang@redhat.com">pahuang@redhat.com</a>
  */
-public class EntityStatusIndexingInterceptor implements
+public class TextFlowTargetEntityIndexingInterceptor implements
         EntityIndexingInterceptor<HTextFlowTarget> {
     @Override
     public IndexingOverride onAdd(HTextFlowTarget hTextFlowTarget) {
--- a/server/zanata-model/src/main/java/org/zanata/model/HTextFlowTarget.java
+++ b/server/zanata-model/src/main/java/org/zanata/model/HTextFlowTarget.java
@@ -60,7 +60,7 @@ import org.zanata.common.ContentState;
 import org.zanata.common.HasContents;
 import org.zanata.common.LocaleId;
 import org.zanata.hibernate.search.ContentStateBridge;
-import org.zanata.hibernate.search.EntityStatusIndexingInterceptor;
+import org.zanata.hibernate.search.TextFlowTargetEntityIndexingInterceptor;
 import org.zanata.hibernate.search.IndexFieldLabels;
 import org.zanata.hibernate.search.LocaleIdBridge;
 import org.zanata.hibernate.search.StringListBridge;
@@ -87,7 +87,7 @@ import org.zanata.model.type.TranslationSourceTypeType;
         @TypeDef(name = "sourceType",
                 typeClass = TranslationSourceTypeType.class),
         @TypeDef(name = "entityType", typeClass = EntityTypeType.class) })
-@Indexed(interceptor = EntityStatusIndexingInterceptor.class)
+@Indexed(interceptor = TextFlowTargetEntityIndexingInterceptor.class)
 public class HTextFlowTarget extends ModelEntityBase
         implements HasContents, HasSimpleComment, ITextFlowTargetHistory,
         Serializable, ITextFlowTarget, IsEntityWithType {
--- a/server/zanata-model/src/main/java/org/zanata/model/HTextFlowTarget.java
+++ b/server/zanata-model/src/main/java/org/zanata/model/HTextFlowTarget.java
@@ -60,6 +60,7 @@ import org.zanata.common.ContentState;
 import org.zanata.common.HasContents;
 import org.zanata.common.LocaleId;
 import org.zanata.hibernate.search.ContentStateBridge;
+import org.zanata.hibernate.search.EntityStatusIndexingInterceptor;
 import org.zanata.hibernate.search.IndexFieldLabels;
 import org.zanata.hibernate.search.LocaleIdBridge;
 import org.zanata.hibernate.search.StringListBridge;
@@ -86,7 +87,7 @@ import org.zanata.model.type.TranslationSourceTypeType;
         @TypeDef(name = "sourceType",
                 typeClass = TranslationSourceTypeType.class),
         @TypeDef(name = "entityType", typeClass = EntityTypeType.class) })
-@Indexed
+@Indexed(interceptor = EntityStatusIndexingInterceptor.class)
 public class HTextFlowTarget extends ModelEntityBase
         implements HasContents, HasSimpleComment, ITextFlowTargetHistory,
         Serializable, ITextFlowTarget, IsEntityWithType {
--- a/server/zanata-war/src/main/java/org/zanata/events/TransMemoryMergeEvent.java
+++ b/server/zanata-war/src/main/java/org/zanata/events/TransMemoryMergeEvent.java
@@ -93,6 +93,6 @@ public class TransMemoryMergeEvent {
     }
 
     public Date getEndTime() {
-        return new Date(endTime.getTime());
+        return endTime == null ? null : new Date(endTime.getTime());
     }
 }
--- a/server/zanata-war/src/main/java/org/zanata/search/HTextFlowTargetIndexingStrategy.java
+++ b/server/zanata-war/src/main/java/org/zanata/search/HTextFlowTargetIndexingStrategy.java
@@ -59,7 +59,7 @@ public class HTextFlowTargetIndexingStrategy
             FullTextSession session) {
         // TODO move this query into something like HTextFlowTargetStreamingDAO
         Query query = session.createQuery(
-                "from HTextFlowTarget tft join fetch tft.locale join fetch tft.textFlow join fetch tft.textFlow.document join fetch tft.textFlow.document.locale join fetch tft.textFlow.document.projectIteration join fetch tft.textFlow.document.projectIteration.project");
+                "from HTextFlowTarget tft join fetch tft.locale join fetch tft.textFlow join fetch tft.textFlow.document join fetch tft.textFlow.document.locale join fetch tft.textFlow.document.projectIteration version join fetch tft.textFlow.document.projectIteration.project project where project.status <> 'O' and version.status <> 'O'");
         query.setFetchSize(Integer.MIN_VALUE);
         return query.scroll(ScrollMode.FORWARD_ONLY);
     }
--- a/server/zanata-war/src/main/java/org/zanata/search/HTextFlowTargetIndexingStrategy.java
+++ b/server/zanata-war/src/main/java/org/zanata/search/HTextFlowTargetIndexingStrategy.java
@@ -27,6 +27,7 @@ import org.hibernate.ScrollMode;
 import org.hibernate.ScrollableResults;
 import org.hibernate.search.FullTextSession;
 import org.zanata.async.AsyncTaskHandle;
+import org.zanata.common.EntityStatus;
 import org.zanata.dao.HTextFlowTargetStreamingDAO;
 import org.zanata.model.HProject;
 import org.zanata.model.HProjectIteration;
@@ -59,8 +60,9 @@ public class HTextFlowTargetIndexingStrategy
             FullTextSession session) {
         // TODO move this query into something like HTextFlowTargetStreamingDAO
         Query query = session.createQuery(
-                "from HTextFlowTarget tft join fetch tft.locale join fetch tft.textFlow join fetch tft.textFlow.document join fetch tft.textFlow.document.locale join fetch tft.textFlow.document.projectIteration version join fetch tft.textFlow.document.projectIteration.project project where project.status <> 'O' and version.status <> 'O'");
-        query.setFetchSize(Integer.MIN_VALUE);
+                "from HTextFlowTarget tft join fetch tft.locale join fetch tft.textFlow join fetch tft.textFlow.document join fetch tft.textFlow.document.locale join fetch tft.textFlow.document.projectIteration version join fetch tft.textFlow.document.projectIteration.project project where project.status <> :state and version.status <> :state");
+        query.setFetchSize(Integer.MIN_VALUE)
+                .setParameter("state", EntityStatus.OBSOLETE);
         return query.scroll(ScrollMode.FORWARD_ONLY);
     }
 
--- a/server/zanata-war/src/main/java/org/zanata/service/impl/EntityListenerUtil.java
+++ b/server/zanata-war/src/main/java/org/zanata/service/impl/EntityListenerUtil.java
@@ -45,7 +45,7 @@ final class EntityListenerUtil {
      * @param entityField field name to look up in entity
      * @return looked up index for a given field for the entity
      */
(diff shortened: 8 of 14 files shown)


seanf:
Reviewed 5 of 5 files at r1.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[server/zanata-model/src/main/java/org/zanata/hibernate/search/ContainingWorkspaceBridge.java, line 45 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/362#-KlbK6JTjv1QenfvjinY:-KlbK6JTjv1QenfvjinZ:blbv2ek) ([raw file](https://github.com/zanata/zanata-platform/blob/a6f3b728c6ee184b39b023598d2a19834bf29d2f/server/zanata-model/src/main/java/org/zanata/hibernate/search/ContainingWorkspaceBridge.java#L45)):*
> ```Java
>         addStringField(IndexFieldLabels.PROJECT_FIELD, project.getId().toString(),
>                 luceneDocument, luceneOptions);
>         addStringField(IndexFieldLabels.PROJECT_VERSION_FIELD,
> ```

We should use new names for these new fields to avoid confusion, eg PROJECT_ID_FIELD etc.

---

*[server/zanata-war/src/main/java/org/zanata/search/HTextFlowTargetIndexingStrategy.java, line 62 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/362#-KlbKRWwJNsBJ40sUPQ5:-KlbKRWx38e9-snX0169:b-8ls4az) ([raw file](https://github.com/zanata/zanata-platform/blob/a6f3b728c6ee184b39b023598d2a19834bf29d2f/server/zanata-war/src/main/java/org/zanata/search/HTextFlowTargetIndexingStrategy.java#L62)):*
> ```Java
>         // TODO move this query into something like HTextFlowTargetStreamingDAO
>         Query query = session.createQuery(
>                 "from HTextFlowTarget tft join fetch tft.locale join fetch tft.textFlow join fetch tft.textFlow.document join fetch tft.textFlow.document.locale join fetch tft.textFlow.document.projectIteration version join fetch tft.textFlow.document.projectIteration.project project where project.status <> 'O' and version.status <> 'O'");
> ```

Instead of assuming that 'O' equals obsolete now and forever (which is not documented here), we should pass in the enum as a parameter. We have other queries which do similar things with enums.

---

*[server/zanata-war/src/main/java/org/zanata/search/HTextFlowTargetIndexingStrategy.java, line 62 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/362#-KlbMRCIC89YmY8BZ4nu:-KlbMRCIC89YmY8BZ4nv:b-5gon4d) ([raw file](https://github.com/zanata/zanata-platform/blob/a6f3b728c6ee184b39b023598d2a19834bf29d2f/server/zanata-war/src/main/java/org/zanata/search/HTextFlowTargetIndexingStrategy.java#L62)):*
> ```Java
>         // TODO move this query into something like HTextFlowTargetStreamingDAO
>         Query query = session.createQuery(
>                 "from HTextFlowTarget tft join fetch tft.locale join fetch tft.textFlow join fetch tft.textFlow.document join fetch tft.textFlow.document.locale join fetch tft.textFlow.document.projectIteration version join fetch tft.textFlow.document.projectIteration.project project where project.status <> 'O' and version.status <> 'O'");
> ```

We need to make sure that a full re-index won't bring back these obsolete records. I think we should be using conditional indexing with an `org.hibernate.search.indexes.interceptor.EntityIndexingInterceptor`:

See https://stackoverflow.com/questions/14602029/hibernate-search-how-to-exclude-entity-from-index

---

*[server/zanata-war/src/main/java/org/zanata/service/impl/SlugEntityUpdatedListener.java, line 74 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/362#-KlbJxOwxTG8kES2feqH:-KlbJxOwxTG8kES2feqI:bib3rcw) ([raw file](https://github.com/zanata/zanata-platform/blob/a6f3b728c6ee184b39b023598d2a19834bf29d2f/server/zanata-war/src/main/java/org/zanata/service/impl/SlugEntityUpdatedListener.java#L74)):*
> ```Java
> 
>     private void reindexIfStatusBecameObsolete(HProjectIteration iteration) {
>         if (iteration.getStatus() == EntityStatus.OBSOLETE) {
> ```

We should reindex if this flag changes in either direction, and change the method name to suit. I'm not sure the class name is still accurate either.

---

*[server/zanata-war/src/main/java/org/zanata/service/impl/SlugEntityUpdatedListener.java, line 88 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/362#-KlbJrL175xTbJ5kyXHH:-KlbJrL175xTbJ5kyXHI:bipaojq) ([raw file](https://github.com/zanata/zanata-platform/blob/a6f3b728c6ee184b39b023598d2a19834bf29d2f/server/zanata-war/src/main/java/org/zanata/service/impl/SlugEntityUpdatedListener.java#L88)):*
> ```Java
> 
>     private void reindexIfStatusBecameObsolete(HProject project) {
>         if (project.getStatus() == EntityStatus.OBSOLETE) {
> ```

We should reindex if this flag changes in either direction, and change the method name to suit.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/362)*

huangp:
Reviewed 1 of 5 files at r1.
Review status: all files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[server/zanata-model/src/main/java/org/zanata/hibernate/search/ContainingWorkspaceBridge.java, line 45 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/362#-KlbK6JTjv1QenfvjinY:-KlpcUt-9rQGdSuR3TXF:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/a6f3b728c6ee184b39b023598d2a19834bf29d2f/server/zanata-model/src/main/java/org/zanata/hibernate/search/ContainingWorkspaceBridge.java#L45)):*

We should use new names for these new fields to avoid confusion, eg PROJECT_ID_FIELD etc.
</blockquote></details>

Done.

---

*[server/zanata-war/src/main/java/org/zanata/search/HTextFlowTargetIndexingStrategy.java, line 62 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/362#-KlbKRWwJNsBJ40sUPQ5:-KlpXaMvs8yk2ZHugzEW:bn503o1) ([raw file](https://github.com/zanata/zanata-platform/blob/a6f3b728c6ee184b39b023598d2a19834bf29d2f/server/zanata-war/src/main/java/org/zanata/search/HTextFlowTargetIndexingStrategy.java#L62)):*

Instead of assuming that 'O' equals obsolete now and forever (which is not documented here), we should pass in the enum as a parameter. We have other queries which do similar things with enums.
</blockquote></details>

We have mixture of both type of queries. Some even use 0 ,1 for enum values directly in query. At least this 'O' thing is harder to change. But I am happy to use the query parameter.

---

*[server/zanata-war/src/main/java/org/zanata/search/HTextFlowTargetIndexingStrategy.java, line 62 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/362#-KlbMRCIC89YmY8BZ4nu:-Klp_E6smvKOkC7uoSaY:b-fakf67) ([raw file](https://github.com/zanata/zanata-platform/blob/a6f3b728c6ee184b39b023598d2a19834bf29d2f/server/zanata-war/src/main/java/org/zanata/search/HTextFlowTargetIndexingStrategy.java#L62)):*

We need to make sure that a full re-index won't bring back these obsolete records. I think we should be using conditional indexing with an `org.hibernate.search.indexes.interceptor.EntityIndexingInterceptor`:

See https://stackoverflow.com/questions/14602029/hibernate-search-how-to-exclude-entity-from-index
</blockquote></details>

We use a specific indexing strategy HTextFlowTargetIndexingStrategy instead of the default simple class indexing strategy. So we shouldn't need to use conditional indexing interceptor anymore. We might use it for HAccount, HProject and HProjectIteration if we want to exclude certain entities from index.

---

*[server/zanata-war/src/main/java/org/zanata/service/impl/SlugEntityUpdatedListener.java, line 74 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/362#-KlbJxOwxTG8kES2feqH:-KlpcjhGYLmtZU2o22-9:b-guqaa3) ([raw file](https://github.com/zanata/zanata-platform/blob/a6f3b728c6ee184b39b023598d2a19834bf29d2f/server/zanata-war/src/main/java/org/zanata/service/impl/SlugEntityUpdatedListener.java#L74)):*

We should reindex if this flag changes in either direction, and change the method name to suit. I'm not sure the class name is still accurate either.
</blockquote></details>

Until we implement ZNTA-309, there is no code path to reach the other direction. For now it will become dead code. 

---

*[server/zanata-war/src/main/java/org/zanata/service/impl/SlugEntityUpdatedListener.java, line 88 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/362#-KlbJrL175xTbJ5kyXHH:-Klp_mYrVWknNyH1eLvp:b-guqaa3) ([raw file](https://github.com/zanata/zanata-platform/blob/a6f3b728c6ee184b39b023598d2a19834bf29d2f/server/zanata-war/src/main/java/org/zanata/service/impl/SlugEntityUpdatedListener.java#L88)):*

We should reindex if this flag changes in either direction, and change the method name to suit.
</blockquote></details>

Until we implement ZNTA-309, there is no code path to reach the other direction. For now it will become dead code. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/362)*

seanf:
Reviewed 3 of 3 files at r2.
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[server/zanata-model/src/main/java/org/zanata/hibernate/search/IndexFieldLabels.java, line 10 at r2](https://reviewable.io:443/reviews/zanata/zanata-platform/362#-KlppLxmsRTqfW2PmtPH:-KlppLxmsRTqfW2PmtPI:b-z0hx87) ([raw file](https://github.com/zanata/zanata-platform/blob/875ed4bfdda474d94fe38f2a7b694aeb22d8ac50/server/zanata-model/src/main/java/org/zanata/hibernate/search/IndexFieldLabels.java#L10)):*
> ```Java
>  */
> public interface IndexFieldLabels {
>     public static final String PROJECT_ID_FIELD = "project";
> ```

We should also change the name of the Lucene field.

---

*[server/zanata-model/src/main/java/org/zanata/hibernate/search/IndexFieldLabels.java, line 38 at r2](https://reviewable.io:443/reviews/zanata/zanata-platform/362#-KlppQ3u-06euJoZ2Ubv:-KlppQ3u-06euJoZ2Ubw:b-z0hx87) ([raw file](https://github.com/zanata/zanata-platform/blob/875ed4bfdda474d94fe38f2a7b694aeb22d8ac50/server/zanata-model/src/main/java/org/zanata/hibernate/search/IndexFieldLabels.java#L38)):*
> ```Java
>     public static final String GLOSSARY_QUALIFIED_NAME = "glossaryEntry.glossary.qualifiedName";
> 
>     String PROJECT_VERSION_ID_FIELD = "projectVersion";
> ```

We should also change the name of the Lucene field.

---

*[server/zanata-war/src/main/java/org/zanata/search/HTextFlowTargetIndexingStrategy.java, line 62 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/362#-KlbKRWwJNsBJ40sUPQ5:-KlprbZzScjqz6QBGtRO:b-w3um0) ([raw file](https://github.com/zanata/zanata-platform/blob/a6f3b728c6ee184b39b023598d2a19834bf29d2f/server/zanata-war/src/main/java/org/zanata/search/HTextFlowTargetIndexingStrategy.java#L62)):*

We have mixture of both type of queries. Some even use 0 ,1 for enum values directly in query. At least this 'O' thing is harder to change. But I am happy to use the query parameter.
</blockquote></details>

Assuming 0 and 1 for true and false might be okay, but anything more than that could be a problem (especially enum ordinals).  Let's try to make this one, at least, a bit safer and more readable.

---

*[server/zanata-war/src/main/java/org/zanata/search/HTextFlowTargetIndexingStrategy.java, line 62 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/362#-KlbMRCIC89YmY8BZ4nu:-KlpsL_ASETqafXWBY1x:b-4wtnlr) ([raw file](https://github.com/zanata/zanata-platform/blob/a6f3b728c6ee184b39b023598d2a19834bf29d2f/server/zanata-war/src/main/java/org/zanata/search/HTextFlowTargetIndexingStrategy.java#L62)):*

We use a specific indexing strategy HTextFlowTargetIndexingStrategy instead of the default simple class indexing strategy. So we shouldn't need to use conditional indexing interceptor anymore. We might use it for HAccount, HProject and HProjectIteration if we want to exclude certain entities from index.
</blockquote></details>

HTextFlowTargetIndexingStrategy is part of the indexing scheme used by our SearchIndexManager, but it isn't understood by Hibernate Search's built-in indexing, like auto-indexing when an entity is created/updated/deleted, or MassIndexer (which we don't use yet).

This means that HTextFlowTargets in obsolete documents would be incorrectly added to the index if (a) they change for some reason or (b) we start using MassIndexer for faster re-indexing.

There's a simple example of an EntityIndexingInterceptor for soft deletes here:

http://in.relation.to/2012/02/29/hibernate-search-41-cr-1-soft-deletes/

---

*[server/zanata-war/src/main/java/org/zanata/service/impl/SlugEntityUpdatedListener.java, line 74 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/362#-KlbJxOwxTG8kES2feqH:-Klpq5EM2OA85DeRk_eb:bamg5ra) ([raw file](https://github.com/zanata/zanata-platform/blob/a6f3b728c6ee184b39b023598d2a19834bf29d2f/server/zanata-war/src/main/java/org/zanata/service/impl/SlugEntityUpdatedListener.java#L74)):*

Until we implement ZNTA-309, there is no code path to reach the other direction. For now it will become dead code. 
</blockquote></details>

Not if we pass in the old and new values, the way we used to pass in the old and new slug.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/362)*

huangp:
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[server/zanata-model/src/main/java/org/zanata/hibernate/search/IndexFieldLabels.java, line 10 at r2](https://reviewable.io:443/reviews/zanata/zanata-platform/362#-KlppLxmsRTqfW2PmtPH:-KlqHj3_tAHv4LI_vDCi:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/875ed4bfdda474d94fe38f2a7b694aeb22d8ac50/server/zanata-model/src/main/java/org/zanata/hibernate/search/IndexFieldLabels.java#L10)):*

We should also change the name of the Lucene field.
</blockquote></details>

Done.

---

*[server/zanata-model/src/main/java/org/zanata/hibernate/search/IndexFieldLabels.java, line 38 at r2](https://reviewable.io:443/reviews/zanata/zanata-platform/362#-KlppQ3u-06euJoZ2Ubv:-KlqHimtbE6wbh6rMUj2:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/875ed4bfdda474d94fe38f2a7b694aeb22d8ac50/server/zanata-model/src/main/java/org/zanata/hibernate/search/IndexFieldLabels.java#L38)):*

We should also change the name of the Lucene field.
</blockquote></details>

Done.

---

*[server/zanata-war/src/main/java/org/zanata/search/HTextFlowTargetIndexingStrategy.java, line 62 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/362#-KlbKRWwJNsBJ40sUPQ5:-KlqHk5ZLTUpXho4FuG_:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/a6f3b728c6ee184b39b023598d2a19834bf29d2f/server/zanata-war/src/main/java/org/zanata/search/HTextFlowTargetIndexingStrategy.java#L62)):*

Assuming 0 and 1 for true and false might be okay, but anything more than that could be a problem (especially enum ordinals).  Let's try to make this one, at least, a bit safer and more readable.
</blockquote></details>

Done.

---

*[server/zanata-war/src/main/java/org/zanata/search/HTextFlowTargetIndexingStrategy.java, line 62 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/362#-KlbMRCIC89YmY8BZ4nu:-KlqHl3UwQjT-O1wdoTF:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/a6f3b728c6ee184b39b023598d2a19834bf29d2f/server/zanata-war/src/main/java/org/zanata/search/HTextFlowTargetIndexingStrategy.java#L62)):*

HTextFlowTargetIndexingStrategy is part of the indexing scheme used by our SearchIndexManager, but it isn't understood by Hibernate Search's built-in indexing, like auto-indexing when an entity is created/updated/deleted, or MassIndexer (which we don't use yet).

This means that HTextFlowTargets in obsolete documents would be incorrectly added to the index if (a) they change for some reason or (b) we start using MassIndexer for faster re-indexing.

There's a simple example of an EntityIndexingInterceptor for soft deletes here:

http://in.relation.to/2012/02/29/hibernate-search-41-cr-1-soft-deletes/
</blockquote></details>

Done.

---

*[server/zanata-war/src/main/java/org/zanata/service/impl/SlugEntityUpdatedListener.java, line 74 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/362#-KlbJxOwxTG8kES2feqH:-KlqHQf0vVXQtv0e6IoA:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/a6f3b728c6ee184b39b023598d2a19834bf29d2f/server/zanata-war/src/main/java/org/zanata/service/impl/SlugEntityUpdatedListener.java#L74)):*

Not if we pass in the old and new values, the way we used to pass in the old and new slug.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/362)*

seanf:
Reviewed 5 of 5 files at r3.
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[server/zanata-model/src/main/java/org/zanata/hibernate/search/EntityStatusIndexingInterceptor.java, line 32 at r3](https://reviewable.io:443/reviews/zanata/zanata-platform/362#-KlqPlv8D6BgTzHsuXVg:-KlqPlv8D6BgTzHsuXVh:b-xg7hr3) ([raw file](https://github.com/zanata/zanata-platform/blob/daf35ce94b2dbeed8ea2523a89a2b17b0541e887/server/zanata-model/src/main/java/org/zanata/hibernate/search/EntityStatusIndexingInterceptor.java#L32)):*
> ```Java
>  * @author Patrick Huang <a href="mailto:pahuang@redhat.com">pahuang@redhat.com</a>
>  */
> public class EntityStatusIndexingInterceptor implements
> ```

I would suggest something like TextFlowTargetEntityIndexingInterceptor, since it is specific to one entity class.

---

*[server/zanata-war/src/main/java/org/zanata/service/impl/SlugEntityUpdatedListener.java, line 34 at r3](https://reviewable.io:443/reviews/zanata/zanata-platform/362#-KlqQgh23qeSIUnFbsyK:-KlqQgh3t8Hrlh5sRjEH:bnfrwvs) ([raw file](https://github.com/zanata/zanata-platform/blob/daf35ce94b2dbeed8ea2523a89a2b17b0541e887/server/zanata-war/src/main/java/org/zanata/service/impl/SlugEntityUpdatedListener.java#L34)):*
> ```Java
>  *         <a href="mailto:pahuang@redhat.com">pahuang@redhat.com</a>
>  */
> public class SlugEntityUpdatedListener implements PostUpdateEventListener {
> ```

I think we should rename this class, since it's not just about slugs any more. I would suggest something like `ReindexPostUpdateEventListener` or `ReindexChildrenEntityEventListener`, but it really performs two different tasks for two different entities, because it also fires events when the slugs change. It would be better to handle one task, or one entity; either would make it easier to find a good name (SRP).

Perhaps we would be better off with standard JPA entity listeners per-class, as used in the EntityListeners for HTextFlow and HTextFlowTarget, instead of Hibernate's PostUpdateEventListener. Unfortunately, that would require changing the implementation, and some transient properties to hold old values, but at least we wouldn't have to make assumptions about the order of properties in the state array being consistent.

---

*[server/zanata-war/src/main/java/org/zanata/service/impl/SlugEntityUpdatedListener.java, line 85 at r3](https://reviewable.io:443/reviews/zanata/zanata-platform/362#-KlqVtouvnYeUvBLoI1y:-KlqVtov82Qe8YD1TpQe:bfjue35) ([raw file](https://github.com/zanata/zanata-platform/blob/daf35ce94b2dbeed8ea2523a89a2b17b0541e887/server/zanata-war/src/main/java/org/zanata/service/impl/SlugEntityUpdatedListener.java#L85)):*
> ```Java
>     private void reindexIfStatusChange(HProjectIteration iteration, EntityStatus oldStatus) {
>         if (iteration.getStatus() == EntityStatus.OBSOLETE
>                 || oldStatus == EntityStatus.OBSOLETE) {
> ```

Couldn't you just use `iteration.getStatus() != oldStatus`?

---

*[server/zanata-war/src/main/java/org/zanata/service/impl/SlugEntityUpdatedListener.java, line 101 at r3](https://reviewable.io:443/reviews/zanata/zanata-platform/362#-KlqW0WgItHB54ErTyvF:-KlqW0WgItHB54ErTyvG:b-uqsz89) ([raw file](https://github.com/zanata/zanata-platform/blob/daf35ce94b2dbeed8ea2523a89a2b17b0541e887/server/zanata-war/src/main/java/org/zanata/service/impl/SlugEntityUpdatedListener.java#L101)):*
> ```Java
>             EntityStatus oldStatus) {
>         if (project.getStatus() == EntityStatus.OBSOLETE
>                 || oldStatus == EntityStatus.OBSOLETE) {
> ```

Again, couldn't you just use `iteration.getStatus() != oldStatus`?

---

*[server/zanata-war/src/main/java/org/zanata/service/impl/SlugEntityUpdatedListener.java, line 163 at r3](https://reviewable.io:443/reviews/zanata/zanata-platform/362#-KlqWG0k1BA-AECFqkVO:-KlqWG0k1BA-AECFqkVP:b-ge0olp) ([raw file](https://github.com/zanata/zanata-platform/blob/daf35ce94b2dbeed8ea2523a89a2b17b0541e887/server/zanata-war/src/main/java/org/zanata/service/impl/SlugEntityUpdatedListener.java#L163)):*
> ```Java
>                 event.getEntity(), Lists.newArrayList(propertyNames));
>         throw new IllegalStateException(
>                 "can not find slug index in entity properties");
> ```

slug should be $entityField

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/362)*

huangp:
Review status: 5 of 11 files reviewed at latest revision, 6 unresolved discussions.

---

*[server/zanata-model/src/main/java/org/zanata/hibernate/search/EntityStatusIndexingInterceptor.java, line 32 at r3](https://reviewable.io:443/reviews/zanata/zanata-platform/362#-KlqPlv8D6BgTzHsuXVg:-KlqerdXpdKKsFu4BSBs:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/daf35ce94b2dbeed8ea2523a89a2b17b0541e887/server/zanata-model/src/main/java/org/zanata/hibernate/search/EntityStatusIndexingInterceptor.java#L32)):*

I would suggest something like TextFlowTargetEntityIndexingInterceptor, since it is specific to one entity class.
</blockquote></details>

Done.

---

*[server/zanata-war/src/main/java/org/zanata/service/impl/SlugEntityUpdatedListener.java, line 34 at r3](https://reviewable.io:443/reviews/zanata/zanata-platform/362#-KlqQgh23qeSIUnFbsyK:-Klqg2ZGV3Wrmli44P5I:b-w9vbcn) ([raw file](https://github.com/zanata/zanata-platform/blob/daf35ce94b2dbeed8ea2523a89a2b17b0541e887/server/zanata-war/src/main/java/org/zanata/service/impl/SlugEntityUpdatedListener.java#L34)):*

I think we should rename this class, since it's not just about slugs any more. I would suggest something like `ReindexPostUpdateEventListener` or `ReindexChildrenEntityEventListener`, but it really performs two different tasks for two different entities, because it also fires events when the slugs change. It would be better to handle one task, or one entity; either would make it easier to find a good name (SRP).

Perhaps we would be better off with standard JPA entity listeners per-class, as used in the EntityListeners for HTextFlow and HTextFlowTarget, instead of Hibernate's PostUpdateEventListener. Unfortunately, that would require changing the implementation, and some transient properties to hold old values, but at least we wouldn't have to make assumptions about the order of properties in the state array being consistent.
</blockquote></details>

Those standard event listeners do not support post commit event (yes they was both called PostUpdateEventListener). This Event Listener is attached to post commit event in HibernateIntegrator. Because we don't want to re-index when the transaction is not yet committed.
Also unlike the standard annotation driven listener, this listener listens to all entity events not just a single type. 

In fact, when I implemented this it was hibernate 4.something. Since 4.3 they introduced some new features. Looks like this code is not doing what I think it would do. Changed and split into two classes.

---

*[server/zanata-war/src/main/java/org/zanata/service/impl/SlugEntityUpdatedListener.java, line 85 at r3](https://reviewable.io:443/reviews/zanata/zanata-platform/362#-KlqVtouvnYeUvBLoI1y:-KlqhRwtaugn3EaCKRZH:b-8q3q7v) ([raw file](https://github.com/zanata/zanata-platform/blob/daf35ce94b2dbeed8ea2523a89a2b17b0541e887/server/zanata-war/src/main/java/org/zanata/service/impl/SlugEntityUpdatedListener.java#L85)):*

Couldn't you just use `iteration.getStatus() != oldStatus`?
</blockquote></details>

no because we don't want to do anything if it's from active to read only (or vise versa)

---

*[server/zanata-war/src/main/java/org/zanata/service/impl/SlugEntityUpdatedListener.java, line 101 at r3](https://reviewable.io:443/reviews/zanata/zanata-platform/362#-KlqW0WgItHB54ErTyvF:-Klqhbgwk4urweWhmdOQ:b-5kwr88) ([raw file](https://github.com/zanata/zanata-platform/blob/daf35ce94b2dbeed8ea2523a89a2b17b0541e887/server/zanata-war/src/main/java/org/zanata/service/impl/SlugEntityUpdatedListener.java#L101)):*

Again, couldn't you just use `iteration.getStatus() != oldStatus`?
</blockquote></details>

same above

---

*[server/zanata-war/src/main/java/org/zanata/service/impl/SlugEntityUpdatedListener.java, line 163 at r3](https://reviewable.io:443/reviews/zanata/zanata-platform/362#-KlqWG0k1BA-AECFqkVO:-KlqhcxlrZYKyBa6scQK:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/daf35ce94b2dbeed8ea2523a89a2b17b0541e887/server/zanata-war/src/main/java/org/zanata/service/impl/SlugEntityUpdatedListener.java#L163)):*

slug should be $entityField
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/362)*

seanf:
Reviewed 7 of 7 files at r4.
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[server/zanata-war/src/main/java/org/zanata/service/impl/EntityListenerUtil.java, line 48 at r4](https://reviewable.io:443/reviews/zanata/zanata-platform/362#-Klr6copXjU6YuvfTuhu:-Klr6copXjU6YuvfTuhv:bx58vv5) ([raw file](https://github.com/zanata/zanata-platform/blob/eba23f3c6960881e58c633ce7af1636cc6a79fc8/server/zanata-war/src/main/java/org/zanata/service/impl/EntityListenerUtil.java#L48)):*
> ```Java
>      * @return looked up index for a given field for the entity
>      */
>     static Integer getFieldIndex(Integer slugFieldIndex,
> ```

Returning `int` would be a little clearer (since it's apparently not nullable).

---

*[server/zanata-war/src/main/java/org/zanata/service/impl/ReindexChildrenEntityEventListener.java, line 68 at r4](https://reviewable.io:443/reviews/zanata/zanata-platform/362#-Klr7zXzzrUZrKRqXBuf:-Klr7zXzzrUZrKRqXBug:brnu33w) ([raw file](https://github.com/zanata/zanata-platform/blob/eba23f3c6960881e58c633ce7af1636cc6a79fc8/server/zanata-war/src/main/java/org/zanata/service/impl/ReindexChildrenEntityEventListener.java#L68)):*
> ```Java
>         if (!entityClass.equals(HProject.class)
>                 && !entityClass.equals(HProjectIteration.class)) {
>             return;
> ```

Do you need to check for these exact classes? What if a future version of Hibernate decides to subclass our entities? This will stop intercepting them without even a log message.

Why not just let the following if-else do its thing with instanceof? (Same applies to the other EventListener, I suppose.)

---

*[server/zanata-war/src/main/java/org/zanata/service/impl/SlugEntityUpdatedListener.java, line 34 at r3](https://reviewable.io:443/reviews/zanata/zanata-platform/362#-KlqQgh23qeSIUnFbsyK:-Klr6yTI94BM63snagL0:bik3dwv) ([raw file](https://github.com/zanata/zanata-platform/blob/daf35ce94b2dbeed8ea2523a89a2b17b0541e887/server/zanata-war/src/main/java/org/zanata/service/impl/SlugEntityUpdatedListener.java#L34)):*

Those standard event listeners do not support post commit event (yes they was both called PostUpdateEventListener). This Event Listener is attached to post commit event in HibernateIntegrator. Because we don't want to re-index when the transaction is not yet committed.
Also unlike the standard annotation driven listener, this listener listens to all entity events not just a single type. 

In fact, when I implemented this it was hibernate 4.something. Since 4.3 they introduced some new features. Looks like this code is not doing what I think it would do. Changed and split into two classes.

</blockquote></details>

Okay, fair enough.

---

*[server/zanata-war/src/main/java/org/zanata/service/impl/SlugEntityUpdatedListener.java, line 71 at r4](https://reviewable.io:443/reviews/zanata/zanata-platform/362#-Klr7KPo0rn2Ii7f0u65:-Klr7KPo0rn2Ii7f0u66:be1zp7c) ([raw file](https://github.com/zanata/zanata-platform/blob/eba23f3c6960881e58c633ce7af1636cc6a79fc8/server/zanata-war/src/main/java/org/zanata/service/impl/SlugEntityUpdatedListener.java#L71)):*
> ```Java
>     @Override
>     public boolean requiresPostCommitHanding(EntityPersister persister) {
>         return true;
> ```

Does this mean you know what this flag means, despite the zero documentation on the parent interface? Does it change *when* the hook is called? Perhaps we could compensate for the lack of docs with a comment here (and in the other class).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/362)*

huangp:
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[server/zanata-war/src/main/java/org/zanata/service/impl/EntityListenerUtil.java, line 48 at r4](https://reviewable.io:443/reviews/zanata/zanata-platform/362#-Klr6copXjU6YuvfTuhu:-KlukQX_YsOreZpzPZYH:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/eba23f3c6960881e58c633ce7af1636cc6a79fc8/server/zanata-war/src/main/java/org/zanata/service/impl/EntityListenerUtil.java#L48)):*

Returning `int` would be a little clearer (since it's apparently not nullable).
</blockquote></details>

Done.

---

*[server/zanata-war/src/main/java/org/zanata/service/impl/ReindexChildrenEntityEventListener.java, line 68 at r4](https://reviewable.io:443/reviews/zanata/zanata-platform/362#-Klr7zXzzrUZrKRqXBuf:-KlukZW9xepymuWATOso:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/eba23f3c6960881e58c633ce7af1636cc6a79fc8/server/zanata-war/src/main/java/org/zanata/service/impl/ReindexChildrenEntityEventListener.java#L68)):*

Do you need to check for these exact classes? What if a future version of Hibernate decides to subclass our entities? This will stop intercepting them without even a log message.

Why not just let the following if-else do its thing with instanceof? (Same applies to the other EventListener, I suppose.)
</blockquote></details>

Done.

---

*[server/zanata-war/src/main/java/org/zanata/service/impl/SlugEntityUpdatedListener.java, line 71 at r4](https://reviewable.io:443/reviews/zanata/zanata-platform/362#-Klr7KPo0rn2Ii7f0u65:-KlujoFNrAN4Ctu06_0j:bjydtt6) ([raw file](https://github.com/zanata/zanata-platform/blob/eba23f3c6960881e58c633ce7af1636cc6a79fc8/server/zanata-war/src/main/java/org/zanata/service/impl/SlugEntityUpdatedListener.java#L71)):*

Does this mean you know what this flag means, despite the zero documentation on the parent interface? Does it change *when* the hook is called? Perhaps we could compensate for the lack of docs with a comment here (and in the other class).
</blockquote></details>

https://dzone.com/articles/yet-another-robust-way-of-listening-hibernate-even
By checking the code, PostCommitEventListener is a subclass of PostUpdateEventListener. This flag distinguish the two. So if you really want post commit (not post update), you return true. I have put in the comment. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/362)*

seanf:
:lgtm:

---

Reviewed 3 of 3 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/362#-:-KluwL8AI6jItg18dSln:bnfp4nl)*

PR #398, opened 2017-06-16
ZNTA-2057 async refactor

huangp:
JIRA issue URL: https://zanata.atlassian.net/browse/ZNTA-2057

* refactor async key from Serializable to Object
* added REST endpoint for querying async tasks (http://localhost:8080/rest/process?includeFinished=true)

## Reviewer tasks

The following is based on the
to refresh your memory.

 - Code quality
   - If code is removed
 - Testing
   - [ ] New code has tests
   - [ ] Changed code has tests
   - [ ] All tests pass
 - Documentation

----

---
This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-platform/398)

### DIFF
--- a/api/zanata-common-api/src/main/java/org/zanata/rest/dto/ProcessStatus.java
+++ b/api/zanata-common-api/src/main/java/org/zanata/rest/dto/ProcessStatus.java
@@ -65,6 +65,10 @@ public class ProcessStatus {
         /** The process has finshed with a failure */
         @XmlEnumValue("Failed")
         Failed,
+
+        /** The process has been cancelled */
+        @XmlEnumValue("Cancelled")
+        Cancelled
     }
 
     private String url;
@@ -109,6 +113,11 @@ public class ProcessStatus {
 
     public void setMessages(List<String> messages) {
         this.messages = messages;
+    }
+
+    public ProcessStatus addMessage(String message) {
+        getMessages().add(message);
+        return this;
     }
 
     @XmlElement
--- a/client/zanata-client-commands/src/main/java/org/zanata/client/commands/push/PushCommand.java
+++ b/client/zanata-client-commands/src/main/java/org/zanata/client/commands/push/PushCommand.java
@@ -601,6 +601,10 @@ public class PushCommand extends PushPullCommand<PushOptions> {
                         ConsoleUtils
                                 .setProgressFeedbackMessage("Waiting for other clients ...");
                         break;
+                    case Cancelled:
+                        waitForCompletion = false;
+                        ConsoleUtils.setProgressFeedbackMessage("Process is cancelled");
+                        break;
                 }
 
                 // Wait before retrying
@@ -724,6 +728,10 @@ public class PushCommand extends PushPullCommand<PushOptions> {
                                                 getOpts().isMyTrans());
                         ConsoleUtils
                                 .setProgressFeedbackMessage("Waiting for other clients ...");
+                        break;
+                    case Cancelled:
+                        waitForCompletion = false;
+                        ConsoleUtils.setProgressFeedbackMessage("Process is cancelled");
                         break;
                 }
 
--- a/server/gwt-editor/src/main/java/org/zanata/webtrans/client/view/DocumentListDisplay.java
+++ b/server/gwt-editor/src/main/java/org/zanata/webtrans/client/view/DocumentListDisplay.java
@@ -86,7 +86,7 @@ public interface DocumentListDisplay extends WidgetDisplay, SearchFieldListener 
 
     void hideConfirmation();
 
-    void updateFileDownloadProgress(int currentProgress, int maxProgress);
+    void updateFileDownloadProgress(long currentProgress, long maxProgress);
 
     void setDownloadInProgress(boolean inProgress);
 
--- a/server/gwt-editor/src/main/java/org/zanata/webtrans/client/view/DocumentListView.java
+++ b/server/gwt-editor/src/main/java/org/zanata/webtrans/client/view/DocumentListView.java
@@ -222,7 +222,7 @@ public class DocumentListView extends Composite implements DocumentListDisplay {
 
     @Override
     public void
-            updateFileDownloadProgress(int currentProgress, int maxProgress) {
+            updateFileDownloadProgress(long currentProgress, long maxProgress) {
         confirmationBox.setProgressMessage(currentProgress + " of "
                 + maxProgress);
     }
--- a/server/gwt-editor/src/main/java/org/zanata/webtrans/shared/rpc/GetDownloadAllFilesProgressResult.java
+++ b/server/gwt-editor/src/main/java/org/zanata/webtrans/shared/rpc/GetDownloadAllFilesProgressResult.java
@@ -4,26 +4,26 @@ public class GetDownloadAllFilesProgressResult implements DispatchResult {
 
     private static final long serialVersionUID = 1L;
 
-    private int currentProgress;
-    private int maxProgress;
+    private long currentProgress;
+    private long maxProgress;
     private String downloadId;
 
     @SuppressWarnings("unused")
     private GetDownloadAllFilesProgressResult() {
     }
 
-    public GetDownloadAllFilesProgressResult(int currentProgress,
-            int maxProgress, String downloadId) {
+    public GetDownloadAllFilesProgressResult(long currentProgress,
+            long maxProgress, String downloadId) {
         this.maxProgress = maxProgress;
         this.currentProgress = currentProgress;
         this.downloadId = downloadId;
     }
 
-    public int getCurrentProgress() {
+    public long getCurrentProgress() {
         return currentProgress;
     }
 
-    public int getMaxProgress() {
+    public long getMaxProgress() {
         return maxProgress;
     }
 
--- a/server/zanata-war/src/main/java/org/zanata/action/CopyTransManager.java
+++ b/server/zanata-war/src/main/java/org/zanata/action/CopyTransManager.java
@@ -149,7 +149,7 @@ public class CopyTransManager implements Serializable {
      * Internal class to index Copy Trans processes.
      */
     private static final class CopyTransProcessKey implements
-            AsyncTaskHandleManager.AsyncTaskKey<CopyTransProcessKey> {
+            AsyncTaskHandleManager.AsyncTaskKey {
         private static final long serialVersionUID = -2054359069473618887L;
         private static final String KEY_NAME = "copyTransKey";
         private String projectSlug;
--- a/server/zanata-war/src/main/java/org/zanata/action/CopyTransManager.java
+++ b/server/zanata-war/src/main/java/org/zanata/action/CopyTransManager.java
@@ -20,7 +20,7 @@
  */
 package org.zanata.action;
 
-import static org.zanata.async.AsyncTaskHandleManager.AsyncTaskKey.joinFields;
+import static org.zanata.async.AsyncTaskKey.joinFields;
 
 import java.io.Serializable;
 
@@ -30,6 +30,8 @@ import javax.inject.Inject;
 
 import org.zanata.async.AsyncTaskHandle;
 import org.zanata.async.AsyncTaskHandleManager;
+import org.zanata.async.AsyncTaskKey;
+import org.zanata.async.GenericAsyncTaskKey;
 import org.zanata.async.handle.CopyTransTaskHandle;
 import org.zanata.model.HCopyTransOptions;
 import org.zanata.model.HDocument;
@@ -63,7 +65,7 @@ public class CopyTransManager implements Serializable {
     private ZanataIdentity identity;
 
     public boolean isCopyTransRunning(@Nonnull Object target) {
-        AsyncTaskHandleManager.AsyncTaskKey key;
+        AsyncTaskKey key;
         if (target instanceof HProjectIteration) {
             key = CopyTransProcessKey.getKey((HProjectIteration) target);
         } else if (target instanceof HDocument) {
@@ -97,7 +99,7 @@ public class CopyTransManager implements Serializable {
                     "Copy Trans is already running for document \'"
                             + document.getDocId() + "\'");
         }
-        AsyncTaskHandleManager.AsyncTaskKey key = CopyTransProcessKey.getKey(document);
+        AsyncTaskKey key = CopyTransProcessKey.getKey(document);
         CopyTransTaskHandle handle = new CopyTransTaskHandle();
         asyncTaskHandleManager.registerTaskHandle(handle, key);
         copyTransServiceImpl.startCopyTransForDocument(document, options,
@@ -115,7 +117,7 @@ public class CopyTransManager implements Serializable {
                     "Copy Trans is already running for version \'"
                             + iteration.getSlug() + "\'");
         }
-        AsyncTaskHandleManager.AsyncTaskKey key = CopyTransProcessKey.getKey(iteration);
+        AsyncTaskKey key = CopyTransProcessKey.getKey(iteration);
         CopyTransTaskHandle handle = new CopyTransTaskHandle();
         asyncTaskHandleManager.registerTaskHandle(handle, key);
         copyTransServiceImpl.startCopyTransForIteration(iteration, options,
@@ -124,7 +126,7 @@ public class CopyTransManager implements Serializable {
 
     public CopyTransTaskHandle
             getCopyTransProcessHandle(@Nonnull Object target) {
-        AsyncTaskHandleManager.AsyncTaskKey key;
+        AsyncTaskKey key;
         if (target instanceof HProjectIteration) {
             key = CopyTransProcessKey.getKey((HProjectIteration) target);
         } else if (target instanceof HDocument) {
@@ -153,16 +155,16 @@ public class CopyTransManager implements Serializable {
         private static final String KEY_NAME = "copyTransKey";
 
 
-        public static AsyncTaskHandleManager.AsyncTaskKey getKey(HProjectIteration iteration) {
-            return new AsyncTaskHandleManager.GenericKey(joinFields(iteration.getProject().getSlug(),
+        public static AsyncTaskKey getKey(HProjectIteration iteration) {
+            return new GenericAsyncTaskKey(joinFields(iteration.getProject().getSlug(),
                     iteration.getSlug(), null));
         }
 
-        public static AsyncTaskHandleManager.AsyncTaskKey getKey(HDocument document) {
+        public static AsyncTaskKey getKey(HDocument document) {
             String projectSlug = document.getProjectIteration().getProject().getSlug();
             String versionSlug = document.getProjectIteration().getSlug();
             String docId = document.getDocId();
-            return new AsyncTaskHandleManager.GenericKey(joinFields(KEY_NAME, projectSlug, versionSlug, docId));
+            return new GenericAsyncTaskKey(joinFields(KEY_NAME, projectSlug, versionSlug, docId));
         }
 
     }
--- a/server/zanata-war/src/main/java/org/zanata/action/CopyTransManager.java
+++ b/server/zanata-war/src/main/java/org/zanata/action/CopyTransManager.java
@@ -203,20 +203,5 @@ public class CopyTransManager implements Serializable {
         public String id() {
             return joinFields(KEY_NAME, projectSlug, iterationSlug, docId);
         }
-
-        @Override
-        public boolean equals(Object o) {
-            if (this == o) return true;
-            if (o == null || getClass() != o.getClass()) return false;
-            CopyTransProcessKey that = (CopyTransProcessKey) o;
-            return Objects.equals(projectSlug, that.projectSlug) &&
-                    Objects.equals(iterationSlug, that.iterationSlug) &&
-                    Objects.equals(docId, that.docId);
-        }
-
-        @Override
-        public int hashCode() {
-            return Objects.hash(projectSlug, iterationSlug, docId);
-        }
     }
 }
--- a/server/zanata-war/src/main/java/org/zanata/action/CopyTransManager.java
+++ b/server/zanata-war/src/main/java/org/zanata/action/CopyTransManager.java
@@ -23,7 +23,6 @@ package org.zanata.action;
 import static org.zanata.async.AsyncTaskHandleManager.AsyncTaskKey.joinFields;
 
 import java.io.Serializable;
-import java.util.Objects;
 
 import javax.annotation.Nonnull;
 import javax.enterprise.context.Dependent;
@@ -64,7 +63,7 @@ public class CopyTransManager implements Serializable {
     private ZanataIdentity identity;
 
     public boolean isCopyTransRunning(@Nonnull Object target) {
-        CopyTransProcessKey key;
+        AsyncTaskHandleManager.AsyncTaskKey key;
         if (target instanceof HProjectIteration) {
             key = CopyTransProcessKey.getKey((HProjectIteration) target);
         } else if (target instanceof HDocument) {
@@ -98,7 +97,7 @@ public class CopyTransManager implements Serializable {
                     "Copy Trans is already running for document \'"
                             + document.getDocId() + "\'");
         }
-        CopyTransProcessKey key = CopyTransProcessKey.getKey(document);
+        AsyncTaskHandleManager.AsyncTaskKey key = CopyTransProcessKey.getKey(document);
         CopyTransTaskHandle handle = new CopyTransTaskHandle();
         asyncTaskHandleManager.registerTaskHandle(handle, key);
         copyTransServiceImpl.startCopyTransForDocument(document, options,
@@ -116,7 +115,7 @@ public class CopyTransManager implements Serializable {
                     "Copy Trans is already running for version \'"
                             + iteration.getSlug() + "\'");
         }
-        CopyTransProcessKey key = CopyTransProcessKey.getKey(iteration);
+        AsyncTaskHandleManager.AsyncTaskKey key = CopyTransProcessKey.getKey(iteration);
         CopyTransTaskHandle handle = new CopyTransTaskHandle();
         asyncTaskHandleManager.registerTaskHandle(handle, key);
         copyTransServiceImpl.startCopyTransForIteration(iteration, options,
@@ -125,7 +124,7 @@ public class CopyTransManager implements Serializable {
 
     public CopyTransTaskHandle
             getCopyTransProcessHandle(@Nonnull Object target) {
-        CopyTransProcessKey key;
+        AsyncTaskHandleManager.AsyncTaskKey key;
         if (target instanceof HProjectIteration) {
             key = CopyTransProcessKey.getKey((HProjectIteration) target);
         } else if (target instanceof HDocument) {
@@ -150,44 +149,21 @@ public class CopyTransManager implements Serializable {
     /**
      * Internal class to index Copy Trans processes.
      */
-    private static final class CopyTransProcessKey extends
-            AsyncTaskHandleManager.GenericKey {
-        private static final long serialVersionUID = -2054359069473618887L;
+    private static final class CopyTransProcessKey {
         private static final String KEY_NAME = "copyTransKey";
-        private final String projectSlug;
-        private final String iterationSlug;
-        private final String docId;
 
-        CopyTransProcessKey(String projectSlug,
-                String iterationSlug, String docId) {
-            super(joinFields(KEY_NAME, projectSlug, iterationSlug, docId));
-            this.projectSlug = projectSlug;
-            this.iterationSlug = iterationSlug;
-            this.docId = docId;
+
+        public static AsyncTaskHandleManager.AsyncTaskKey getKey(HProjectIteration iteration) {
+            return new AsyncTaskHandleManager.GenericKey(joinFields(iteration.getProject().getSlug(),
+                    iteration.getSlug(), null));
         }
 
-        public static CopyTransProcessKey getKey(HProjectIteration iteration) {
-            return new CopyTransProcessKey(iteration.getProject().getSlug(),
-                    iteration.getSlug(), null);
-        }
-
-        public static CopyTransProcessKey getKey(HDocument document) {
+        public static AsyncTaskHandleManager.AsyncTaskKey getKey(HDocument document) {
             String projectSlug = document.getProjectIteration().getProject().getSlug();
             String versionSlug = document.getProjectIteration().getSlug();
             String docId = document.getDocId();
-            return new CopyTransProcessKey(projectSlug, versionSlug, docId);
+            return new AsyncTaskHandleManager.GenericKey(joinFields(KEY_NAME, projectSlug, versionSlug, docId));
         }
 
-        public String getProjectSlug() {
-            return this.projectSlug;
-        }
-
-        public String getIterationSlug() {
-            return this.iterationSlug;
-        }
-
-        public String getDocId() {
-            return this.docId;
-        }
     }
 }
--- a/server/zanata-war/src/main/java/org/zanata/action/CopyTransManager.java
+++ b/server/zanata-war/src/main/java/org/zanata/action/CopyTransManager.java
@@ -21,7 +21,7 @@
 package org.zanata.action;
 
 import java.io.Serializable;
-import java.util.List;
+import java.util.Objects;
 
 import javax.annotation.Nonnull;
 import javax.enterprise.context.Dependent;
@@ -35,8 +35,6 @@ import org.zanata.model.HDocument;
 import org.zanata.model.HProjectIteration;
 import org.zanata.security.ZanataIdentity;
 import org.zanata.service.CopyTransService;
-
-import com.google.common.base.Preconditions;
 
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 // TODO This class should be merged with the copy trans service (?)
@@ -207,13 +205,18 @@ public class CopyTransManager implements Serializable {
         }
 
         @Override
-        public CopyTransProcessKey from(String id) {
-            List<String> parts = parseId(id, KEY_NAME, 3);
-            CopyTransProcessKey key = new CopyTransProcessKey();
-            key.projectSlug = parts.get(0);
-            key.iterationSlug = parts.get(1);
-            key.docId = parts.get(2);
-            return key;
+        public boolean equals(Object o) {
+            if (this == o) return true;
+            if (o == null || getClass() != o.getClass()) return false;
+            CopyTransProcessKey that = (CopyTransProcessKey) o;
+            return Objects.equals(projectSlug, that.projectSlug) &&
+                    Objects.equals(iterationSlug, that.iterationSlug) &&
+                    Objects.equals(docId, that.docId);
+        }
+
+        @Override
+        public int hashCode() {
+            return Objects.hash(projectSlug, iterationSlug, docId);
         }
     }
 }
--- a/server/zanata-war/src/main/java/org/zanata/action/CopyTransManager.java
+++ b/server/zanata-war/src/main/java/org/zanata/action/CopyTransManager.java
@@ -203,5 +203,18 @@ public class CopyTransManager implements Serializable {
         public String id() {
             return joinFields(KEY_NAME, projectSlug, iterationSlug, docId);
         }
+
+        @Override
+        public boolean equals(Object o) {
+            if (this == o) return true;
+            if (o == null || getClass() != o.getClass()) return false;
+            CopyTransProcessKey that = (CopyTransProcessKey) o;
+            return Objects.equals(id(), that.id());
+        }
+
+        @Override
+        public int hashCode() {
+            return Objects.hash(id());
+        }
     }
 }
--- a/server/zanata-war/src/main/java/org/zanata/action/CopyTransManager.java
+++ b/server/zanata-war/src/main/java/org/zanata/action/CopyTransManager.java
@@ -20,6 +20,8 @@
  */
 package org.zanata.action;
 
+import static org.zanata.async.AsyncTaskHandleManager.AsyncTaskKey.joinFields;
+
 import java.io.Serializable;
 import java.util.Objects;
 
@@ -148,28 +150,32 @@ public class CopyTransManager implements Serializable {
     /**
      * Internal class to index Copy Trans processes.
      */
-    private static final class CopyTransProcessKey implements
-            AsyncTaskHandleManager.AsyncTaskKey {
+    private static final class CopyTransProcessKey extends
+            AsyncTaskHandleManager.GenericKey {
         private static final long serialVersionUID = -2054359069473618887L;
         private static final String KEY_NAME = "copyTransKey";
-        private String projectSlug;
-        private String iterationSlug;
-        private String docId;
+        private final String projectSlug;
+        private final String iterationSlug;
+        private final String docId;
+
+        CopyTransProcessKey(String projectSlug,
+                String iterationSlug, String docId) {
+            super(joinFields(KEY_NAME, projectSlug, iterationSlug, docId));
+            this.projectSlug = projectSlug;
+            this.iterationSlug = iterationSlug;
+            this.docId = docId;
+        }
 
         public static CopyTransProcessKey getKey(HProjectIteration iteration) {
-            CopyTransProcessKey newKey = new CopyTransProcessKey();
-            newKey.setProjectSlug(iteration.getProject().getSlug());
-            newKey.setIterationSlug(iteration.getSlug());
-            return newKey;
+            return new CopyTransProcessKey(iteration.getProject().getSlug(),
+                    iteration.getSlug(), null);
         }
 
         public static CopyTransProcessKey getKey(HDocument document) {
-            CopyTransProcessKey newKey = new CopyTransProcessKey();
-            newKey.setDocId(document.getDocId());
-            newKey.setProjectSlug(
-                    document.getProjectIteration().getProject().getSlug());
-            newKey.setIterationSlug(document.getProjectIteration().getSlug());
-            return newKey;
+            String projectSlug = document.getProjectIteration().getProject().getSlug();
+            String versionSlug = document.getProjectIteration().getSlug();
+            String docId = document.getDocId();
+            return new CopyTransProcessKey(projectSlug, versionSlug, docId);
         }
 
         public String getProjectSlug() {
@@ -182,39 +188,6 @@ public class CopyTransManager implements Serializable {
 
         public String getDocId() {
             return this.docId;
-        }
-
-        public void setProjectSlug(final String projectSlug) {
-            this.projectSlug = projectSlug;
-        }
-
-        public void setIterationSlug(final String iterationSlug) {
-            this.iterationSlug = iterationSlug;
-        }
-
-        public void setDocId(final String docId) {
-            this.docId = docId;
-        }
-
-        private CopyTransProcessKey() {
-        }
-
-        @Override
-        public String id() {
-            return joinFields(KEY_NAME, projectSlug, iterationSlug, docId);
-        }
-
-        @Override
-        public boolean equals(Object o) {
-            if (this == o) return true;
-            if (o == null || getClass() != o.getClass()) return false;
-            CopyTransProcessKey that = (CopyTransProcessKey) o;
-            return Objects.equals(id(), that.id());
-        }
-
-        @Override
-        public int hashCode() {
-            return Objects.hash(id());
         }
     }
 }
--- a/server/zanata-war/src/main/java/org/zanata/action/CopyTransManager.java
+++ b/server/zanata-war/src/main/java/org/zanata/action/CopyTransManager.java
@@ -101,6 +101,7 @@ public class CopyTransManager implements Serializable {
         }
         AsyncTaskKey key = CopyTransProcessKey.getKey(document);
         CopyTransTaskHandle handle = new CopyTransTaskHandle();
+        handle.setTriggeredBy(identity.getAccountUsername());
         asyncTaskHandleManager.registerTaskHandle(handle, key);
         copyTransServiceImpl.startCopyTransForDocument(document, options,
                 handle);
@@ -119,6 +120,7 @@ public class CopyTransManager implements Serializable {
         }
         AsyncTaskKey key = CopyTransProcessKey.getKey(iteration);
         CopyTransTaskHandle handle = new CopyTransTaskHandle();
+        handle.setTriggeredBy(identity.getAccountUsername());
         asyncTaskHandleManager.registerTaskHandle(handle, key);
         copyTransServiceImpl.startCopyTransForIteration(iteration, options,
                 handle);
--- a/server/zanata-war/src/main/java/org/zanata/action/CopyTransManager.java
+++ b/server/zanata-war/src/main/java/org/zanata/action/CopyTransManager.java
@@ -21,11 +21,12 @@
 package org.zanata.action;
 
 import java.io.Serializable;
+import java.util.List;
+
 import javax.annotation.Nonnull;
-import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import javax.enterprise.context.Dependent;
 import javax.inject.Inject;
-import javax.inject.Named;
+
 import org.zanata.async.AsyncTaskHandle;
 import org.zanata.async.AsyncTaskHandleManager;
 import org.zanata.async.handle.CopyTransTaskHandle;
@@ -34,6 +35,10 @@ import org.zanata.model.HDocument;
 import org.zanata.model.HProjectIteration;
 import org.zanata.security.ZanataIdentity;
 import org.zanata.service.CopyTransService;
+
+import com.google.common.base.Preconditions;
+
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 // TODO This class should be merged with the copy trans service (?)
 
 /**
@@ -145,8 +150,10 @@ public class CopyTransManager implements Serializable {
     /**
      * Internal class to index Copy Trans processes.
      */
-    private static final class CopyTransProcessKey implements Serializable {
+    private static final class CopyTransProcessKey implements
+            AsyncTaskHandleManager.AsyncTaskKey<CopyTransProcessKey> {
         private static final long serialVersionUID = -2054359069473618887L;
+        private static final String KEY_NAME = "copyTransKey";
         private String projectSlug;
         private String iterationSlug;
         private String docId;
@@ -165,46 +172,6 @@ public class CopyTransManager implements Serializable {
                     document.getProjectIteration().getProject().getSlug());
             newKey.setIterationSlug(document.getProjectIteration().getSlug());
             return newKey;
-        }
-
-        @Override
-        public boolean equals(final Object o) {
-            if (o == this)
-                return true;
-            if (!(o instanceof CopyTransManager.CopyTransProcessKey))
-                return false;
-            final CopyTransProcessKey other = (CopyTransProcessKey) o;
-            final Object this$projectSlug = this.getProjectSlug();
-            final Object other$projectSlug = other.getProjectSlug();
-            if (this$projectSlug == null ? other$projectSlug != null
-                    : !this$projectSlug.equals(other$projectSlug))
-                return false;
-            final Object this$iterationSlug = this.getIterationSlug();
-            final Object other$iterationSlug = other.getIterationSlug();
-            if (this$iterationSlug == null ? other$iterationSlug != null
-                    : !this$iterationSlug.equals(other$iterationSlug))
-                return false;
-            final Object this$docId = this.getDocId();
-            final Object other$docId = other.getDocId();
-            if (this$docId == null ? other$docId != null
-                    : !this$docId.equals(other$docId))
-                return false;
-            return true;
-        }
-
-        @Override
-        public int hashCode() {
-            final int PRIME = 59;
-            int result = 1;
-            final Object $projectSlug = this.getProjectSlug();
-            result = result * PRIME
-                    + ($projectSlug == null ? 43 : $projectSlug.hashCode());
-            final Object $iterationSlug = this.getIterationSlug();
-            result = result * PRIME
-                    + ($iterationSlug == null ? 43 : $iterationSlug.hashCode());
-            final Object $docId = this.getDocId();
-            result = result * PRIME + ($docId == null ? 43 : $docId.hashCode());
-            return result;
         }
 
         public String getProjectSlug() {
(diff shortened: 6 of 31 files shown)


seanf:
Reviewed 25 of 26 files at r1.
Review status: 25 of 26 files reviewed at latest revision, 16 unresolved discussions.

---

*[server/zanata-war/src/main/java/org/zanata/async/AsyncTaskHandleManager.java, line 94 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-KmxzMm_6Ler0WSxRn1a:-KmxzMmaXlcRG9rbf-5K:b-291wtf) ([raw file](https://github.com/zanata/zanata-platform/blob/82fe345ae6382062a0439d8b4adf189196de57ca/server/zanata-war/src/main/java/org/zanata/async/AsyncTaskHandleManager.java#L94)):*
> ```Java
>             return handlesByKey.get(key);
>         }
>         return finishedTasks.getIfPresent(key.id());
> ```

Can we use the Key itself as a, well, key? Might help with the O(n) problems in this class.

We might also want to consider having a single key class (various factory methods, but just one class) so that it's easy to turn an id string back into a Key object.

---

*[server/zanata-war/src/main/java/org/zanata/async/AsyncTaskHandleManager.java, line 139 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-KmxyBlUGdsRP9NefuFq:-KmxyBlUGdsRP9NefuFr:b-9i2afj) ([raw file](https://github.com/zanata/zanata-platform/blob/82fe345ae6382062a0439d8b4adf189196de57ca/server/zanata-war/src/main/java/org/zanata/async/AsyncTaskHandleManager.java#L139)):*
> ```Java
>         String SEPARATOR = "-";
> 
>         String id();
> ```

It might be worth adding a default implementation for toString, which could call id.

---

*[server/zanata-war/src/main/java/org/zanata/async/UserTriggerableTaskHandle.java, line 23 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kmy1b7HzmOztRQa5hQg:-Kmy1b7IZnCQCDnPxb8k:bn03v6g) ([raw file](https://github.com/zanata/zanata-platform/blob/82fe345ae6382062a0439d8b4adf189196de57ca/server/zanata-war/src/main/java/org/zanata/async/UserTriggerableTaskHandle.java#L23)):*
> ```Java
> package org.zanata.async;
> 
> public interface UserTriggerableTaskHandle {
> ```

Seems more like CancellableTask?

---

*[server/zanata-war/src/main/java/org/zanata/async/handle/TransMemoryMergeTaskHandle.java, line 61 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-KmxxXlFcFsfZl0Kk4q-:-KmxxXlFcFsfZl0Kk4q0:b-t8j750) ([raw file](https://github.com/zanata/zanata-platform/blob/82fe345ae6382062a0439d8b4adf189196de57ca/server/zanata-war/src/main/java/org/zanata/async/handle/TransMemoryMergeTaskHandle.java#L61)):*
> ```Java
>                 .add("mergeTarget", mergeTarget)
>                 .add("textFlowFilled", currentProgress)
>                 .add("totalTextFlows", maxProgress)
> ```

These changes make it unclear what currentProgress/maxProgress mean for this type of task. Perhaps we should keep around the old getters/setters and have them delegate to get/set currentProgress/maxProgress. (Or some other solution.)

---

*[server/zanata-war/src/main/java/org/zanata/async/handle/TransMemoryMergeTaskHandle.java, line 62 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-KmxxGKfH7nR1C9JwRzR:-KmxxGKgcbsnLFEA0V3e:b5rm3mw) ([raw file](https://github.com/zanata/zanata-platform/blob/82fe345ae6382062a0439d8b4adf189196de57ca/server/zanata-war/src/main/java/org/zanata/async/handle/TransMemoryMergeTaskHandle.java#L62)):*
> ```Java
>                 .add("textFlowFilled", currentProgress)
>                 .add("totalTextFlows", maxProgress)
>                 .add("cancelledBy", getTriggeredBy())
> ```

This assumes that the canceller is always the user who triggered the task. This will be wrong if we let admin cancel user tasks. Perhaps you should add a method getCancelledBy() which calls getTriggeredBy() for now.

---

*[server/zanata-war/src/main/java/org/zanata/rest/service/AsynchronousProcessResourceService.java, line 117 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kmy-SWTiX27AkcauivW:-Kmy-SWTiX27AkcauivX:be1vn17) ([raw file](https://github.com/zanata/zanata-platform/blob/82fe345ae6382062a0439d8b4adf189196de57ca/server/zanata-war/src/main/java/org/zanata/rest/service/AsynchronousProcessResourceService.java#L117)):*
> ```Java
>                         resource, extensions, copytrans, true, handle);
>         logWhenUploadComplete(handle, name, taskKey.id());
>         return getProcessStatus(taskKey.id()); // TODO Change to return 202
> ```

Why not pass in the key itself to these methods (and to AsyncTaskHandleManager)?

---

*[server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java, line 100 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kmy36qbzz-njmfomdK_:-Kmy36qbzz-njmfomdKa:b-e4vkl8) ([raw file](https://github.com/zanata/zanata-platform/blob/82fe345ae6382062a0439d8b4adf189196de57ca/server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java#L100)):*
> ```Java
>         ProcessStatus status = handleToProcessStatus(handle,
>                 uriInfo.getRequestUri().toString());
>         return Response.ok(status).build();
> ```

We should do an authorisation check.

---

*[server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java, line 104 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kmy09TlxYGtm5zyTHSb:-Kmy09TlxYGtm5zyTHSc:bnqktim) ([raw file](https://github.com/zanata/zanata-platform/blob/82fe345ae6382062a0439d8b4adf189196de57ca/server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java#L104)):*
> ```Java
> 
>     /**
>      * Get all async task's status.
> ```

Get all async task's status. -> Get all async tasks' statuses. (or Get status of/for all async tasks.)

Is this for all users, or just the current user? If the latter, seems like an admin-only function.

---

*[server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java, line 118 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kmy0dolz5KAikFua7Le:-Kmy0dolz5KAikFua7Lf:b-c2f4k6) ([raw file](https://github.com/zanata/zanata-platform/blob/82fe345ae6382062a0439d8b4adf189196de57ca/server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java#L118)):*
> ```Java
>      */
>     @GET
>     public Response getAllAsyncProcess(
> ```

The other method has "status" in the name.

---

*[server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java, line 155 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kmy1M9tQF-muxil6xnt:-Kmy1M9tQF-muxil6xnu:baaxfd7) ([raw file](https://github.com/zanata/zanata-platform/blob/82fe345ae6382062a0439d8b4adf189196de57ca/server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java#L155)):*
> ```Java
>             return Response.status(Response.Status.NOT_FOUND).build();
>         }
>         if (handle.isStarted() && !handle.isCancelled() && !handle.isDone()) {
> ```

Can we encapsulate that boolean logic into a single method somewhere?

---

*[server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java, line 159 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kmy1kUkGV8xWKOx4cvC:-Kmy1kUkGV8xWKOx4cvD:bl1xgph) ([raw file](https://github.com/zanata/zanata-platform/blob/82fe345ae6382062a0439d8b4adf189196de57ca/server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java#L159)):*
> ```Java
>                 UserTriggerableTaskHandle taskHandle =
>                         (UserTriggerableTaskHandle) handle;
>                 if (!taskHandle.canCancel(identity.getAccountUsername())) {
> ```

We should try to encapsulate all that logic into the canCancel method.

---

*[server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java, line 161 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kmy0xNapHSVUleDA553:-Kmy0xNapHSVUleDA554:b8pn4yt) ([raw file](https://github.com/zanata/zanata-platform/blob/82fe345ae6382062a0439d8b4adf189196de57ca/server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java#L161)):*
> ```Java
>                 if (!taskHandle.canCancel(identity.getAccountUsername())) {
>                     throw new AuthorizationException(
>                             "Only the task triggerer or admin can cancel the task.");
> ```

"Triggerer" sounds a bit unwieldy. Perhaps "owner" or "creator" would sound better?

---

*[server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java, line 165 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kmy27J29PqVu2sL-cCI:-Kmy27J29PqVu2sL-cCJ:bb3od56) ([raw file](https://github.com/zanata/zanata-platform/blob/82fe345ae6382062a0439d8b4adf189196de57ca/server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java#L165)):*
> ```Java
>             }
> 
>             handle.cancel(true);
> ```

Does this mean that some types of task can be cancelled by *any* authenticated user? That doesn't seem right.

---

*[server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java, line 182 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kmy2pXqp60pfJ1TA7Cz:-Kmy2pXqp60pfJ1TA7D-:bcr8its) ([raw file](https://github.com/zanata/zanata-platform/blob/82fe345ae6382062a0439d8b4adf189196de57ca/server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java#L182)):*
> ```Java
>                 handle.isDone() ? ProcessStatus.ProcessStatusCode.Finished
>                         : ProcessStatus.ProcessStatusCode.Running);
>         int perComplete = 100;
> ```

percentComplete would be better. Or just percentage.

---

*[server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java, line 191 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kmy3MwzSuAIr5W9twcg:-Kmy3MwzSuAIr5W9twch:b-owod2t) ([raw file](https://github.com/zanata/zanata-platform/blob/82fe345ae6382062a0439d8b4adf189196de57ca/server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java#L191)):*
> ```Java
>         if (handle.isCancelled()) {
>             status.setStatusCode(ProcessStatus.ProcessStatusCode.Cancelled);
>             status.addMessage("Cancelled");
> ```

Do we want to include the username?

---

*[server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java, line 198 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kmy3G88attISjLcQYuc:-Kmy3G88attISjLcQYud:b4qn7mx) ([raw file](https://github.com/zanata/zanata-platform/blob/82fe345ae6382062a0439d8b4adf189196de57ca/server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java#L198)):*
> ```Java
>             } catch (InterruptedException e) {
>                 // The process was forcefully cancelled
>                 status.setStatusCode(ProcessStatus.ProcessStatusCode.Failed);
> ```

Shouldn't this say ProcessStatusCode.Cancelled then? Do we want to include the username?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/398)*

huangp:
Review status: 22 of 26 files reviewed at latest revision, 16 unresolved discussions.

---

*[server/zanata-war/src/main/java/org/zanata/async/AsyncTaskHandleManager.java, line 94 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-KmxzMm_6Ler0WSxRn1a:-Kmy8AoD7Qx3ehWdA4xB:b8jbjny) ([raw file](https://github.com/zanata/zanata-platform/blob/82fe345ae6382062a0439d8b4adf189196de57ca/server/zanata-war/src/main/java/org/zanata/async/AsyncTaskHandleManager.java#L94)):*

Can we use the Key itself as a, well, key? Might help with the O(n) problems in this class.

We might also want to consider having a single key class (various factory methods, but just one class) so that it's easy to turn an id string back into a Key object.
</blockquote></details>

Fixed. In my first commit I actually attempted to add methods to convert between id and object. But then I realize there is no use case to convert from id to key object (other than saving us from the O(n). Given that we have a thread pool of fixed size so that n is not going to be huge. I didn't bother going that route. 

---

*[server/zanata-war/src/main/java/org/zanata/async/AsyncTaskHandleManager.java, line 139 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-KmxyBlUGdsRP9NefuFq:-Kmy7mJNqyAqPPWKyoh4:byrzy4v) ([raw file](https://github.com/zanata/zanata-platform/blob/82fe345ae6382062a0439d8b4adf189196de57ca/server/zanata-war/src/main/java/org/zanata/async/AsyncTaskHandleManager.java#L139)):*

It might be worth adding a default implementation for toString, which could call id.
</blockquote></details>

We can't do that. Java won't allow it.

---

*[server/zanata-war/src/main/java/org/zanata/async/UserTriggerableTaskHandle.java, line 23 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kmy1b7HzmOztRQa5hQg:-Kmy8rDyXGltcLlJdAHR:bc87usz) ([raw file](https://github.com/zanata/zanata-platform/blob/82fe345ae6382062a0439d8b4adf189196de57ca/server/zanata-war/src/main/java/org/zanata/async/UserTriggerableTaskHandle.java#L23)):*

Seems more like CancellableTask?
</blockquote></details>

Well all tasks can be cancelled (by admin). Is just some tasks are triggered by user and should only be cancelled by that user. 

---

*[server/zanata-war/src/main/java/org/zanata/async/handle/TransMemoryMergeTaskHandle.java, line 61 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-KmxxXlFcFsfZl0Kk4q-:-Kmy9E-83lXfSe4HRTUi:b-gogrrr) ([raw file](https://github.com/zanata/zanata-platform/blob/82fe345ae6382062a0439d8b4adf189196de57ca/server/zanata-war/src/main/java/org/zanata/async/handle/TransMemoryMergeTaskHandle.java#L61)):*

These changes make it unclear what currentProgress/maxProgress mean for this type of task. Perhaps we should keep around the old getters/setters and have them delegate to get/set currentProgress/maxProgress. (Or some other solution.)
</blockquote></details>

When I implemented this I didn't realize there are progress fields in the super class. I could just reuse those two fields. The display didn't mention the field names (just a percent), so it's fine to just remove the two fields and use progress fields. I will rename the toString method (only used in debug I believe)

---

*[server/zanata-war/src/main/java/org/zanata/async/handle/TransMemoryMergeTaskHandle.java, line 62 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-KmxxGKfH7nR1C9JwRzR:-Kmy9oXW6ZSBg0eagYiI:bqvohh4) ([raw file](https://github.com/zanata/zanata-platform/blob/82fe345ae6382062a0439d8b4adf189196de57ca/server/zanata-war/src/main/java/org/zanata/async/handle/TransMemoryMergeTaskHandle.java#L62)):*

This assumes that the canceller is always the user who triggered the task. This will be wrong if we let admin cancel user tasks. Perhaps you should add a method getCancelledBy() which calls getTriggeredBy() for now.
</blockquote></details>

Right. In fact the toString is only used in log or debugging. I will get rid of it so It's not confusing. 

---

*[server/zanata-war/src/main/java/org/zanata/rest/service/AsynchronousProcessResourceService.java, line 117 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kmy-SWTiX27AkcauivW:-KmyAKTvAl6bfnHAF7Ir:b1wfhaz) ([raw file](https://github.com/zanata/zanata-platform/blob/82fe345ae6382062a0439d8b4adf189196de57ca/server/zanata-war/src/main/java/org/zanata/rest/service/AsynchronousProcessResourceService.java#L117)):*

Why not pass in the key itself to these methods (and to AsyncTaskHandleManager)?
</blockquote></details>

getProcessStatus is a method in API. Until we deprecate the method (move this to the generic Async REST API), we can't change it.

---

*[server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java, line 100 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kmy36qbzz-njmfomdK_:-KmyB4zblbWpu7uY2e9k:bbmz7jh) ([raw file](https://github.com/zanata/zanata-platform/blob/82fe345ae6382062a0439d8b4adf189196de57ca/server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java#L100)):*

We should do an authorisation check.
</blockquote></details>

Only admin and the task owner can see the process status? Do we need to be that strict for get GET?

---

*[server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java, line 104 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kmy09TlxYGtm5zyTHSb:-KmyBW2_denJGS3RUMwk:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/82fe345ae6382062a0439d8b4adf189196de57ca/server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java#L104)):*

Get all async task's status. -> Get all async tasks' statuses. (or Get status of/for all async tasks.)

Is this for all users, or just the current user? If the latter, seems like an admin-only function.
</blockquote></details>

Done.

---

*[server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java, line 118 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kmy0dolz5KAikFua7Le:-KmyBi5792A-ZUVjtueo:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/82fe345ae6382062a0439d8b4adf189196de57ca/server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java#L118)):*

The other method has "status" in the name.
</blockquote></details>

Done.

---

*[server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java, line 155 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kmy1M9tQF-muxil6xnt:-Kmy546l399ZBLYEUaqF:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/82fe345ae6382062a0439d8b4adf189196de57ca/server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java#L155)):*

Can we encapsulate that boolean logic into a single method somewhere?
</blockquote></details>

Done.

---

*[server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java, line 159 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kmy1kUkGV8xWKOx4cvC:-Kmy4Fnrv45tbiEJj225:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/82fe345ae6382062a0439d8b4adf189196de57ca/server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java#L159)):*

We should try to encapsulate all that logic into the canCancel method.
</blockquote></details>

Done.

---

*[server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java, line 161 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kmy0xNapHSVUleDA553:-Kmy4Xpm-H4cv3q5YHEY:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/82fe345ae6382062a0439d8b4adf189196de57ca/server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java#L161)):*

"Triggerer" sounds a bit unwieldy. Perhaps "owner" or "creator" would sound better?
</blockquote></details>

Done.

---

*[server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java, line 165 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kmy27J29PqVu2sL-cCI:-Kmy4dlyrrDqbJ4g_qXa:bncggmn) ([raw file](https://github.com/zanata/zanata-platform/blob/82fe345ae6382062a0439d8b4adf189196de57ca/server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java#L165)):*

Does this mean that some types of task can be cancelled by *any* authenticated user? That doesn't seem right.
</blockquote></details>

I forgot to implement the other part. For those "system" type task, we should only allow admin to cancel them.

---

*[server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java, line 182 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kmy2pXqp60pfJ1TA7Cz:-Kmy6_y5RNduXtFCIfOk:b-csy3vc) ([raw file](https://github.com/zanata/zanata-platform/blob/82fe345ae6382062a0439d8b4adf189196de57ca/server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java#L182)):*

percentComplete would be better. Or just percentage.
</blockquote></details>

This chunk was moved from other place. Will rename it.

---

*[server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java, line 191 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kmy3MwzSuAIr5W9twcg:-Kmy6pGqd_eMogcwgboL:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/82fe345ae6382062a0439d8b4adf189196de57ca/server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java#L191)):*

Do we want to include the username?
</blockquote></details>

Done.

---

*[server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java, line 198 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kmy3G88attISjLcQYuc:-Kmy6t2dgoRClgzcOA5w:b-xc49jz) ([raw file](https://github.com/zanata/zanata-platform/blob/82fe345ae6382062a0439d8b4adf189196de57ca/server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java#L198)):*

Shouldn't this say ProcessStatusCode.Cancelled then? Do we want to include the username?
</blockquote></details>

I moved this code from AsynchronousProcessResourceService. I found that if you cancel a task, upon handle.getResult(), it will throw a CancellationException instead of InterruptedException. That's why I added the above cancel check block. I am not sure whether this block will even work. Maybe only if you can somehow get hold of the running thread and interrupt that thread (e.g. during a server shutdown?)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/398)*

seanf:
Reviewed 6 of 6 files at r2.
Review status: 24 of 27 files reviewed at latest revision, 9 unresolved discussions.

---

*[server/zanata-war/src/main/java/org/zanata/async/AsyncTaskHandleManager.java, line 94 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-KmxzMm_6Ler0WSxRn1a:-KmypZ6HG_joCqOkr4UR:b-ws49gj) ([raw file](https://github.com/zanata/zanata-platform/blob/82fe345ae6382062a0439d8b4adf189196de57ca/server/zanata-war/src/main/java/org/zanata/async/AsyncTaskHandleManager.java#L94)):*

Fixed. In my first commit I actually attempted to add methods to convert between id and object. But then I realize there is no use case to convert from id to key object (other than saving us from the O(n). Given that we have a thread pool of fixed size so that n is not going to be huge. I didn't bother going that route. 
</blockquote></details>

O(n) is actually pretty terrible, and getting rid of it is a valid use case. Just because the thread pool is limited doesn't really guarantee that n is small.

The only reason we need the O(n) methods is because we keep switching between these String ids and the map key. It's a sign that we are misusing our collections.

This method is called by the CLI (via AsynchronousProcessResource.getProcessStatus) fairly frequently when waiting for an upload to finish, so it's worth choosing good data structures.

If the key string ids are unique anyway, perhaps we should just key the hashmap by string id instead of the so-called Key object.

---

*[server/zanata-war/src/main/java/org/zanata/async/UserTriggerableTaskHandle.java, line 23 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kmy1b7HzmOztRQa5hQg:-KmylZx3j20ko_MrGSJh:b-m5v3l2) ([raw file](https://github.com/zanata/zanata-platform/blob/82fe345ae6382062a0439d8b4adf189196de57ca/server/zanata-war/src/main/java/org/zanata/async/UserTriggerableTaskHandle.java#L23)):*

Well all tasks can be cancelled (by admin). Is just some tasks are triggered by user and should only be cancelled by that user. 
</blockquote></details>

In that case it could probably use more javadoc, because it's still not clear what this really represents.

The reason I suggested that name is that the only method on this interface is about cancelling.

What about UserTriggeredTask? (Perhaps with Handle at the end.)

---

*[server/zanata-war/src/main/java/org/zanata/rest/service/AsynchronousProcessResourceService.java, line 117 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kmy-SWTiX27AkcauivW:-KmyrAErIzAIlFxmhoo2:bmhgf3k) ([raw file](https://github.com/zanata/zanata-platform/blob/82fe345ae6382062a0439d8b4adf189196de57ca/server/zanata-war/src/main/java/org/zanata/rest/service/AsynchronousProcessResourceService.java#L117)):*

getProcessStatus is a method in API. Until we deprecate the method (move this to the generic Async REST API), we can't change it.
</blockquote></details>

Oh, you mean it's in zanata-common-api? We have a lot of API besides zanata-common-api.

Why can't we deprecate it in place and add another? But I'm guessing the various handle/key classes aren't part of zanata-common-api. That's fine.

We could still add an overload here (not declared in zanata-common-api) which can accept a key object instead of the string. And logWhenUploadComplete can take a key object without changing the public API.

Remember that the current getProcessStatus(String) is currently O(n), but getProcessStatus(Key) would be O(1) - unless we switch that hashmap to be keyed by String of course. 

---

*[server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java, line 100 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kmy36qbzz-njmfomdK_:-Kmyqb4w_QVtuSsH4iTv:bqqm6gy) ([raw file](https://github.com/zanata/zanata-platform/blob/82fe345ae6382062a0439d8b4adf189196de57ca/server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java#L100)):*

Only admin and the task owner can see the process status? Do we need to be that strict for get GET?
</blockquote></details>

Some level of strictness in between might be nice, but exposing any handle to any authenticated user (eg regardless of project affiliation) seems like a problem.

---

*[server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java, line 198 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kmy3G88attISjLcQYuc:-KmyojBHV7sycMg6BeU2:be5j6fn) ([raw file](https://github.com/zanata/zanata-platform/blob/82fe345ae6382062a0439d8b4adf189196de57ca/server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java#L198)):*

I moved this code from AsynchronousProcessResourceService. I found that if you cancel a task, upon handle.getResult(), it will throw a CancellationException instead of InterruptedException. That's why I added the above cancel check block. I am not sure whether this block will even work. Maybe only if you can somehow get hold of the running thread and interrupt that thread (e.g. during a server shutdown?)
</blockquote></details>

If we want preemptive cancellation, and at some point I think we will, we will need to make more use of InterruptedException in future. It's the best way to cancel a lot of things.

From an external point of view, whether it was "cancelled" or "interrupted" is an implementation detail which shouldn't be visible. If we're going to have a state for Cancelled (and we should), it should be used whenever a process is cancelled, preemptively or not.

---

*[server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java, line 202 at r2](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kmynr2BhPu3cN8SJEzQ:-Kmynr2CieNID3qTTohl:b8t1bnf) ([raw file](https://github.com/zanata/zanata-platform/blob/ddc82d3fc6a6c5c60144faa09e7fc5243bdbf23e/server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java#L202)):*
> ```Java
>         handle.cancel(true);
>         handle.setCancelledBy(identity.getAccountUsername());
>         handle.setCancelledTime(System.currentTimeMillis());
> ```

Perhaps we should move all the canCancel check, user/admin cancel operation and setCancelled* calls into the handle.cancel method. We should be able to avoid the `if instanceof UserTriggerableTaskHandle` with a bit of polymorphism.

---

*[server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java, line 211 at r2](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kmymh1xvTkitElzlJnf:-Kmymh1y-5p3D24hIRa4:b-tzx6er) ([raw file](https://github.com/zanata/zanata-platform/blob/ddc82d3fc6a6c5c60144faa09e7fc5243bdbf23e/server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java#L211)):*
> ```Java
>     private static boolean taskIsRunning(AsyncTaskHandle handle) {
>         return handle.isStarted() && !handle.isCancelled() && !handle.isDone();
>     }
> ```

This might be better off in the class AsyncTaskHandle.

---

*[server/zanata-war/src/test/java/org/zanata/rest/service/AsyncProcessServiceTest.java, line 88 at r2](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kmytvgmfbii2WMF919q:-KmytvgnYX-nBYgCegnP:b-a9071) ([raw file](https://github.com/zanata/zanata-platform/blob/ddc82d3fc6a6c5c60144faa09e7fc5243bdbf23e/server/zanata-war/src/test/java/org/zanata/rest/service/AsyncProcessServiceTest.java#L88)):*
> ```Java
>         Response response = service.getAsyncProcessStatus("id");
> 
>         assertThat(response.getStatus()).isEqualTo(200);
> ```

We should test a user trying to access someone else's task too. Probably 404.

---

*[server/zanata-war/src/test/java/org/zanata/rest/service/AsyncProcessServiceTest.java, line 131 at r2](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-KmytegQ8RDNZfZ9SOSH:-KmytegQ8RDNZfZ9SOSI:budupjw) ([raw file](https://github.com/zanata/zanata-platform/blob/ddc82d3fc6a6c5c60144faa09e7fc5243bdbf23e/server/zanata-war/src/test/java/org/zanata/rest/service/AsyncProcessServiceTest.java#L131)):*
> ```Java
>         Response response = service.getAllAsyncProcessStatuses(true);
> 
>         assertThat(response.getStatus()).isEqualTo(200);
> ```

We should test a non-admin user for this end-point, and the one above. Probably UNAUTHORIZED.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/398)*

seanf:
Reviewed 3 of 3 files at r3.
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/398)*

huangp:
Review status: 10 of 28 files reviewed at latest revision, 8 unresolved discussions.

---

*[server/zanata-war/src/main/java/org/zanata/async/AsyncTaskHandleManager.java, line 94 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-KmxzMm_6Ler0WSxRn1a:-Kmz2vEi4sxjvkcETQo4:b-8rf7t5) ([raw file](https://github.com/zanata/zanata-platform/blob/82fe345ae6382062a0439d8b4adf189196de57ca/server/zanata-war/src/main/java/org/zanata/async/AsyncTaskHandleManager.java#L94)):*

O(n) is actually pretty terrible, and getting rid of it is a valid use case. Just because the thread pool is limited doesn't really guarantee that n is small.

The only reason we need the O(n) methods is because we keep switching between these String ids and the map key. It's a sign that we are misusing our collections.

This method is called by the CLI (via AsynchronousProcessResource.getProcessStatus) fairly frequently when waiting for an upload to finish, so it's worth choosing good data structures.

If the key string ids are unique anyway, perhaps we should just key the hashmap by string id instead of the so-called Key object.
</blockquote></details>

Done. I think the reason we chose Key object is because we can can have meaningful equal and hashcode implementation as well as toString method. Some tasks are meant to be run at most one at a time for same key (e.g. copy version). Since the id is generated from the field values, I think it's fine to use it as key.

---

*[server/zanata-war/src/main/java/org/zanata/rest/service/AsynchronousProcessResourceService.java, line 117 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kmy-SWTiX27AkcauivW:-Kmz4eljSr4jAPlIlfwE:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/82fe345ae6382062a0439d8b4adf189196de57ca/server/zanata-war/src/main/java/org/zanata/rest/service/AsynchronousProcessResourceService.java#L117)):*

Oh, you mean it's in zanata-common-api? We have a lot of API besides zanata-common-api.

Why can't we deprecate it in place and add another? But I'm guessing the various handle/key classes aren't part of zanata-common-api. That's fine.

We could still add an overload here (not declared in zanata-common-api) which can accept a key object instead of the string. And logWhenUploadComplete can take a key object without changing the public API.

Remember that the current getProcessStatus(String) is currently O(n), but getProcessStatus(Key) would be O(1) - unless we switch that hashmap to be keyed by String of course. 
</blockquote></details>

Done.

---

*[server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java, line 100 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kmy36qbzz-njmfomdK_:-Kmz9H0aiVBjEHYP_zx2:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/82fe345ae6382062a0439d8b4adf189196de57ca/server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java#L100)):*

Some level of strictness in between might be nice, but exposing any handle to any authenticated user (eg regardless of project affiliation) seems like a problem.
</blockquote></details>

Done.

---

*[server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java, line 198 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kmy3G88attISjLcQYuc:-KmzAn3kv6F7fDnDExTX:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/82fe345ae6382062a0439d8b4adf189196de57ca/server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java#L198)):*

If we want preemptive cancellation, and at some point I think we will, we will need to make more use of InterruptedException in future. It's the best way to cancel a lot of things.

From an external point of view, whether it was "cancelled" or "interrupted" is an implementation detail which shouldn't be visible. If we're going to have a state for Cancelled (and we should), it should be used whenever a process is cancelled, preemptively or not.
</blockquote></details>

Done.

---

*[server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java, line 202 at r2](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kmynr2BhPu3cN8SJEzQ:-Kmz9Fbub5ElYkPb5FQL:b-e1mij6) ([raw file](https://github.com/zanata/zanata-platform/blob/ddc82d3fc6a6c5c60144faa09e7fc5243bdbf23e/server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java#L202)):*

Perhaps we should move all the canCancel check, user/admin cancel operation and setCancelled* calls into the handle.cancel method. We should be able to avoid the `if instanceof UserTriggerableTaskHandle` with a bit of polymorphism.
</blockquote></details>

There are some handles that just call the cancel method without setting CancelledBy and CancelledTime. I tried not to make too big a change to the existing code.

---

*[server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java, line 211 at r2](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kmymh1xvTkitElzlJnf:-Kmz74r0Yf4SK-VXzKLr:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/ddc82d3fc6a6c5c60144faa09e7fc5243bdbf23e/server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java#L211)):*

This might be better off in the class AsyncTaskHandle.
</blockquote></details>

Done.

---

*[server/zanata-war/src/test/java/org/zanata/rest/service/AsyncProcessServiceTest.java, line 88 at r2](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kmytvgmfbii2WMF919q:-Kn1iiEsdoYUPNQ0BBgq:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/ddc82d3fc6a6c5c60144faa09e7fc5243bdbf23e/server/zanata-war/src/test/java/org/zanata/rest/service/AsyncProcessServiceTest.java#L88)):*

We should test a user trying to access someone else's task too. Probably 404.
</blockquote></details>

Done.

---

*[server/zanata-war/src/test/java/org/zanata/rest/service/AsyncProcessServiceTest.java, line 131 at r2](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-KmytegQ8RDNZfZ9SOSH:-Kn1xTCvJ7IkgkjK4aLo:bbrrjx8) ([raw file](https://github.com/zanata/zanata-platform/blob/ddc82d3fc6a6c5c60144faa09e7fc5243bdbf23e/server/zanata-war/src/test/java/org/zanata/rest/service/AsyncProcessServiceTest.java#L131)):*

We should test a non-admin user for this end-point, and the one above. Probably UNAUTHORIZED.
</blockquote></details>

This is controlled by CDI interceptor. I added an arquillian test for this.

---

*[server/zanata-war/src/main/java/org/zanata/async/UserTriggerableTaskHandle.java, line 23 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kmy1b7HzmOztRQa5hQg:-Kmz4XusOoiEjlKGNDbO:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/82fe345ae6382062a0439d8b4adf189196de57ca/server/zanata-war/src/main/java/org/zanata/async/UserTriggerableTaskHandle.java#L23)):*

In that case it could probably use more javadoc, because it's still not clear what this really represents.

The reason I suggested that name is that the only method on this interface is about cancelling.

What about UserTriggeredTask? (Perhaps with Handle at the end.)
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/398)*

seanf:
Reviewed 13 of 18 files at r5, 6 of 6 files at r6.
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[server/zanata-war/src/main/java/org/zanata/async/AsyncTaskHandleManager.java, line 62 at r5](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kn1rMojkpDZMr806pgg:-Kn1rMojkpDZMr806pgh:bsa607z) ([raw file](https://github.com/zanata/zanata-platform/blob/e753e86ce8476996501daa1587731bee69e8a585/server/zanata-war/src/main/java/org/zanata/async/AsyncTaskHandleManager.java#L62)):*
> ```Java
>                     + " already exists");
>         }
>         handlesByKey.put(key.id(), handle);
> ```

This could use HashMap.putIfAbsent now, instead of containsKey/put.

---

*[server/zanata-war/src/main/java/org/zanata/async/AsyncTaskHandleManager.java, line 72 at r5](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kn1s9xbDtOjGQM2qVoM:-Kn1s9xbDtOjGQM2qVoN:bsakcy1) ([raw file](https://github.com/zanata/zanata-platform/blob/e753e86ce8476996501daa1587731bee69e8a585/server/zanata-war/src/main/java/org/zanata/async/AsyncTaskHandleManager.java#L72)):*
> ```Java
>     public synchronized String registerTaskHandle(AsyncTaskHandle handle) {
>         String autoGenKey = UUID.randomUUID().toString();
>         AsyncTaskKey autoKey = () -> autoGenKey;
> ```

So we're using AsyncTaskKey like a FunctionalInterface (even though it doesn't have the annotation)? This seems error-prone, because the temptation in this method would be to write:

    AsyncTaskKey autoKey = () -> UUID.randomUUID().toString();

Whereas one of the most important things about a key is that it be immutable.

If it's going to be a FunctionalInterface, perhaps it would be better as an AsyncTaskKeyGen (after all, it just generates the id string which is the real key now).

Otherwise it might be safest if AsyncTaskKey were a final immutable class which just holds a String.

---

*[server/zanata-war/src/main/java/org/zanata/async/AsyncTaskHandleManager.java, line 82 at r5](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kn1uSp9mOSKU4Rq2nMJ:-Kn1uSp9mOSKU4Rq2nMK:b-4iv64e) ([raw file](https://github.com/zanata/zanata-platform/blob/e753e86ce8476996501daa1587731bee69e8a585/server/zanata-war/src/main/java/org/zanata/async/AsyncTaskHandleManager.java#L82)):*
> ```Java
>             for (Map.Entry<String, AsyncTaskHandle<?>> entry : handlesByKey
>                     .entrySet()) {
>                 if (entry.getValue().equals(taskHandle)) {
> ```

We should change AsyncTaskHandle to hold its Key (or the String), so that we can extract it and use it for efficient lookups (O(1) instead of O(n)).

---

*[server/zanata-war/src/main/java/org/zanata/async/AsyncTaskHandleManager.java, line 138 at r5](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kn1uwxQo0FRuaOjIacg:-Kn1uwxQo0FRuaOjIach:bschxed) ([raw file](https://github.com/zanata/zanata-platform/blob/e753e86ce8476996501daa1587731bee69e8a585/server/zanata-war/src/main/java/org/zanata/async/AsyncTaskHandleManager.java#L138)):*
> ```Java
>          * @return String representation of the key which can be used as id
>          */
>         default String joinFields(String keyName, String... fields) {
> ```

This might be better as `static` than `default`.

---

*[server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java, line 202 at r2](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kmynr2BhPu3cN8SJEzQ:-Kn1yX0mysjQ-xyt9QRH:b-ljrjgd) ([raw file](https://github.com/zanata/zanata-platform/blob/ddc82d3fc6a6c5c60144faa09e7fc5243bdbf23e/server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java#L202)):*

There are some handles that just call the cancel method without setting CancelledBy and CancelledTime. I tried not to make too big a change to the existing code.
</blockquote></details>

That sounds like kind of a bug then, which crept in because the API didn't enforce it before. 

We could always have two cancel methods, one of which doesn't require a username but is deprecated, to be called by the old bad code.

---

*[server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java, line 126 at r5](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kn1vDKksifVQ8XTKnBL:-Kn1vDKlp0CoS1AihuiX:bmte5xy) ([raw file](https://github.com/zanata/zanata-platform/blob/e753e86ce8476996501daa1587731bee69e8a585/server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java#L126)):*
> ```Java
>                 String triggeredBy =
>                         ((UserTriggeredTaskHandle) handle).getTriggeredBy();
>                 if (!triggeredBy.equals(identity.getAccountUsername())) {
> ```

If you move this check into, say, `UserTriggeredTaskHandle.visibleTo(String accountName)`, we should be able to change the access rules more cleanly, and share the logic in non-REST parts of the system.

---

*[server/zanata-war/src/test/java/org/zanata/rest/service/AsyncProcessServiceTest.java, line 131 at r2](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-KmytegQ8RDNZfZ9SOSH:-Kn1zJ8Kxc1Q25ptREXK:b4n0k3l) ([raw file](https://github.com/zanata/zanata-platform/blob/ddc82d3fc6a6c5c60144faa09e7fc5243bdbf23e/server/zanata-war/src/test/java/org/zanata/rest/service/AsyncProcessServiceTest.java#L131)):*

This is controlled by CDI interceptor. I added an arquillian test for this.
</blockquote></details>

Okay. We should have some comments explaining why those tests aren't here in this class.

---

*[server/zanata-war/src/test/java/org/zanata/rest/service/AsyncProcessServiceTest.java, line 111 at r5](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kn1vwfbSFXNoSCCwClC:-Kn1vwfbSFXNoSCCwClD:b9ontk8) ([raw file](https://github.com/zanata/zanata-platform/blob/e753e86ce8476996501daa1587731bee69e8a585/server/zanata-war/src/test/java/org/zanata/rest/service/AsyncProcessServiceTest.java#L111)):*
> ```Java
> 
>     @Test
>     public void adminUserCanNotSeeOthersProcessStatus() throws URISyntaxException {
> ```

"Not" doesn't belong in the name.

---

*[server/zanata-war/src/test/java/org/zanata/rest/service/raw/AsyncProcessRawRestITCase.java, line 66 at r5](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kn1xAjXInWmIvehqh2B:-Kn1xAjXInWmIvehqh2C:badmvgd) ([raw file](https://github.com/zanata/zanata-platform/blob/e753e86ce8476996501daa1587731bee69e8a585/server/zanata-war/src/test/java/org/zanata/rest/service/raw/AsyncProcessRawRestITCase.java#L66)):*
> ```Java
>     @Test
>     @RunAsClient
>     public void adminCanGetAllProcessStatuses() throws Exception {
> ```

Doesn't this duplicate one of the tests in AsyncProcessServiceTest?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/398)*

huangp:
Review status: 16 of 30 files reviewed at latest revision, 9 unresolved discussions.

---

*[server/zanata-war/src/main/java/org/zanata/async/AsyncTaskHandleManager.java, line 62 at r5](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kn1rMojkpDZMr806pgg:-Kn2589KqP5HI-sxR6IM:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/e753e86ce8476996501daa1587731bee69e8a585/server/zanata-war/src/main/java/org/zanata/async/AsyncTaskHandleManager.java#L62)):*

This could use HashMap.putIfAbsent now, instead of containsKey/put.
</blockquote></details>

Done.

---

*[server/zanata-war/src/main/java/org/zanata/async/AsyncTaskHandleManager.java, line 72 at r5](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kn1s9xbDtOjGQM2qVoM:-Kn259s9qZxzv_6FQETa:b-t8abe8) ([raw file](https://github.com/zanata/zanata-platform/blob/e753e86ce8476996501daa1587731bee69e8a585/server/zanata-war/src/main/java/org/zanata/async/AsyncTaskHandleManager.java#L72)):*

So we're using AsyncTaskKey like a FunctionalInterface (even though it doesn't have the annotation)? This seems error-prone, because the temptation in this method would be to write:

    AsyncTaskKey autoKey = () -> UUID.randomUUID().toString();

Whereas one of the most important things about a key is that it be immutable.

If it's going to be a FunctionalInterface, perhaps it would be better as an AsyncTaskKeyGen (after all, it just generates the id string which is the real key now).

Otherwise it might be safest if AsyncTaskKey were a final immutable class which just holds a String.

</blockquote></details>

No. We have a bunch of subclasses implementing their own id() method. Those keys can have more semantic meanings of what it represent.  This is the default one where you just want a unique random key (e.g. allowing the same type of process to run at the same time). I will create a GenericKey as subclass so that it's immutable.

---

*[server/zanata-war/src/main/java/org/zanata/async/AsyncTaskHandleManager.java, line 82 at r5](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kn1uSp9mOSKU4Rq2nMJ:-Kn27mTzrzm-2MJvvp3e:b-6wcf0q) ([raw file](https://github.com/zanata/zanata-platform/blob/e753e86ce8476996501daa1587731bee69e8a585/server/zanata-war/src/main/java/org/zanata/async/AsyncTaskHandleManager.java#L82)):*

We should change AsyncTaskHandle to hold its Key (or the String), so that we can extract it and use it for efficient lookups (O(1) instead of O(n)).
</blockquote></details>

Thought about that. I wanted to do incremental change and merge so that this PR won't grow too long and stale. This PR is intended just to fix the Async key and add a new REST api for general control of tasks. Other refactoring or improvement of existing code can come later (this change will touch multiple JSF action class and a few other REST endpoints).

---

*[server/zanata-war/src/main/java/org/zanata/async/AsyncTaskHandleManager.java, line 138 at r5](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kn1uwxQo0FRuaOjIacg:-Kn28apwvnXB5qNrzEgQ:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/e753e86ce8476996501daa1587731bee69e8a585/server/zanata-war/src/main/java/org/zanata/async/AsyncTaskHandleManager.java#L138)):*

This might be better as `static` than `default`.
</blockquote></details>

Done.

---

*[server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java, line 202 at r2](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kmynr2BhPu3cN8SJEzQ:-Kn2IAAX3NHOzz4dK4m9:bqdkzfo) ([raw file](https://github.com/zanata/zanata-platform/blob/ddc82d3fc6a6c5c60144faa09e7fc5243bdbf23e/server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java#L202)):*

That sounds like kind of a bug then, which crept in because the API didn't enforce it before. 

We could always have two cancel methods, one of which doesn't require a username but is deprecated, to be called by the old bad code.
</blockquote></details>

Again as discussed I want to merge often and sooner so not to change too much of existing code. I need to study for each individual task, how the cancellation is triggered (by user or by system) and how those values (cancelledBy and time) are being used. This can come as a separate PR.

---

*[server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java, line 126 at r5](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kn1vDKksifVQ8XTKnBL:-Kn2I7J_wQZrDGYd4ZKb:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/e753e86ce8476996501daa1587731bee69e8a585/server/zanata-war/src/main/java/org/zanata/rest/service/AsyncProcessService.java#L126)):*

If you move this check into, say, `UserTriggeredTaskHandle.visibleTo(String accountName)`, we should be able to change the access rules more cleanly, and share the logic in non-REST parts of the system.
</blockquote></details>

Done.

---

*[server/zanata-war/src/test/java/org/zanata/rest/service/AsyncProcessServiceTest.java, line 131 at r2](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-KmytegQ8RDNZfZ9SOSH:-Kn24-mM-R6Dzay6mjD-:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/ddc82d3fc6a6c5c60144faa09e7fc5243bdbf23e/server/zanata-war/src/test/java/org/zanata/rest/service/AsyncProcessServiceTest.java#L131)):*

Okay. We should have some comments explaining why those tests aren't here in this class.
</blockquote></details>

Done.

---

*[server/zanata-war/src/test/java/org/zanata/rest/service/AsyncProcessServiceTest.java, line 111 at r5](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kn1vwfbSFXNoSCCwClC:-Kn240sdwmYpy981AAoA:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/e753e86ce8476996501daa1587731bee69e8a585/server/zanata-war/src/test/java/org/zanata/rest/service/AsyncProcessServiceTest.java#L111)):*

"Not" doesn't belong in the name.
</blockquote></details>

Done.

---

*[server/zanata-war/src/test/java/org/zanata/rest/service/raw/AsyncProcessRawRestITCase.java, line 66 at r5](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-Kn1xAjXInWmIvehqh2B:-Kn242DMu0fOjswYpX9o:b-6714fz) ([raw file](https://github.com/zanata/zanata-platform/blob/e753e86ce8476996501daa1587731bee69e8a585/server/zanata-war/src/test/java/org/zanata/rest/service/raw/AsyncProcessRawRestITCase.java#L66)):*

Doesn't this duplicate one of the tests in AsyncProcessServiceTest?
</blockquote></details>

This checks the admin permission (go through CDI interceptor). It doesn't set up any real data (the result is empty since nothing is running or finished on server). The one in unit test checks the actual implementation.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/398)*

seanf:
:lgtm:

---

Reviewed 1 of 18 files at r5, 14 of 14 files at r7.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/398#-:-Kn33R_pdklgkJJBKcgc:bnfp4nl)*

djansen-redhat:
Reviewed 10 of 27 files at r1, 7 of 18 files at r5, 12 of 14 files at r7, 2 of 2 files at r8.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/398)*

PR #498, opened 2017-08-25
Switch to wildfly-maven-plugin

seanf:
This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-platform/498)

### DIFF
--- a/build
+++ b/build
@@ -322,7 +322,7 @@ fi
 
 
 #############################################
-# cargo:run (for --wait)
+# wildfly:run (for --wait)
 #############################################
 
 unset cargocommandline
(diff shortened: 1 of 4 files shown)


seanf:
I did have automated tests working, but I didn't find a way of replicating the required behaviour of cargo:run with wildfly:run. From memory, the problem was the inability to deploy (1) the JDBC driver, (2) datasource xml and (3) zanata-test.war, all together with wildfly:run. wildfly:run starts the appserver, but after that it doesn't have enough hooks for deploying everything we need (afterDeployment is deprecated).

Also, with the automated tests, I think there was a danger of leaving the appserver running when aborting a test run (if post-integration-test is not invoked).

PR #504, opened 2017-08-29
fix: add enabled index in HAccount

aeng:
https://zanata.atlassian.net/browse/ZNTA-2191

*[Add a short description of what is changed here.]*

## Reviewer tasks

The following is based on the
to refresh your memory.

 - Code quality
   - If code is removed
 - Testing
   - [ ] New code has tests
   - [ ] Changed code has tests
   - [ ] All tests pass
 - Documentation

----

---
This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-platform/504)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -290,7 +290,7 @@ timestamps {
                         //[parserName: 'appserver log messages'], // 119 warnings
                         //[parserName: 'browser warnings'],       // 0 warnings
                 ],
-                unstableTotalAll: '419',
+                unstableTotalAll: '300',
                 unstableTotalHigh: '0',
           ])
           // TODO check integration test warnings (EAP and WildFly)
@@ -316,7 +316,7 @@ timestamps {
         sh "git clean -fdx"
       } catch (e) {
         echo("Caught exception: " + e)
-        notify.error(e)
+        notify.error(e.toString())
         currentBuild.result = 'FAILURE'
         // abort the rest of the pipeline
         throw e
--- a/api/zanata-common-api/src/main/java/org/zanata/rest/service/LocalesResource.java
+++ b/api/zanata-common-api/src/main/java/org/zanata/rest/service/LocalesResource.java
@@ -12,6 +12,7 @@ import javax.ws.rs.QueryParam;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 
+import com.webcohesion.enunciate.metadata.rs.ResourceLabel;
 import com.webcohesion.enunciate.metadata.rs.ResponseCode;
 import com.webcohesion.enunciate.metadata.rs.StatusCodes;
 import com.webcohesion.enunciate.metadata.rs.TypeHint;
@@ -22,8 +23,14 @@ import org.zanata.rest.dto.SourceLocaleDetails;
 /**
  * @author Alex Eng <a href="mailto:aeng@redhat.com">aeng@redhat.com</a>
  */
+@Path(LocalesResource.SERVICE_PATH)
+@ResourceLabel("Languages")
 @Produces({ MediaType.APPLICATION_JSON })
 @Consumes({ MediaType.APPLICATION_JSON })
+@StatusCodes({
+        @ResponseCode(code = 500,
+                condition = "If there is an unexpected error in the server while performing this operation")
+})
 public interface LocalesResource {
 
     public static final String SERVICE_PATH = "/locales";
@@ -33,9 +40,7 @@ public interface LocalesResource {
      *
      * @return The following response status codes will be returned from this
      *         operation:<br>
-     *         OK(200) - Response containing a full list of locales. <br>
-     *         INTERNAL SERVER ERROR(500) - If there is an unexpected error in
-     *         the server while performing this operation.
+     *         OK(200) - Response containing a full list of locales.
      */
     @GET
     @Produces(MediaType.APPLICATION_JSON)
@@ -67,9 +72,7 @@ public interface LocalesResource {
      *
      * @return The following response status codes will be returned from this
      *         operation:<br>
-     *         OK(200) - Response containing a full list of locales. <br>
-     *         INTERNAL SERVER ERROR(500) - If there is an unexpected error in
-     *         the server while performing this operation.
+     *         OK(200) - Response containing a full list of locales.
      */
     @GET
     @Path("/ui")
@@ -99,9 +102,7 @@ public interface LocalesResource {
      *
      * @return The following response status codes will be returned from this
      *         operation:<br>
-     *         OK(200) - Response containing a list of locales. <br>
-     *         INTERNAL SERVER ERROR(500) - If there is an unexpected error in
-     *         the server while performing this operation.
+     *         OK(200) - Response containing a list of locales.
      */
     @GET
     @Path("/new")
@@ -115,9 +116,7 @@ public interface LocalesResource {
      *
      * @return The following response status codes will be returned from this
      *         operation:<br>
-     *         OK(200) - Locale is deleted. <br>
-     *         INTERNAL SERVER ERROR(500) - If there is an unexpected error in
-     *         the server while performing this operation.
+     *         OK(200) - Locale is deleted.
      */
     @DELETE
     @Path("/locale/{localeId}")
@@ -129,9 +128,7 @@ public interface LocalesResource {
      * @return The following response status codes will be returned from this
      *         operation:<br>
      *         OK(201) - Locale is added. <br>
-     *         OK(200) - Locale is already exist. <br>
-     *         INTERNAL SERVER ERROR(500) - If there is an unexpected error in
-     *         the server while performing this operation.
+     *         OK(200) - Locale is already exist.
      */
     @PUT
     @Path("/locale")
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -1,3 +1,13 @@
+## 4.2.4
+##### Changes
+ * [ZNTA-2035](https://zanata.atlassian.net/browse/ZNTA-2035) - Fix Jenkin's CI reporting to Github
+
+##### Bug Fixes
+ * [ZNTA-2114](https://zanata.atlassian.net/browse/ZNTA-2114) - Online help broken
+ * [ZNTA-128](https://zanata.atlassian.net/browse/ZNTA-128) - Chunk split failed when pushing big files (more than 1MiB)
+
+-----------------------
+
 ## 4.2.0
 ##### Changes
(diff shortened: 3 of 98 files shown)


huangp:
Reviewed 2 of 3 files at r1.
Review status: 2 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

*[server/zanata-model/src/main/java/org/zanata/model/HAccount.java, line 115 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/504#-KskKq0dBpMLRZpwjOsZ:-KskKq0dBpMLRZpwjOs_:bnrtycq) ([raw file](https://github.com/zanata/zanata-platform/blob/55190479cc8e6bef30988d50af3c4474d0b8a715/server/zanata-model/src/main/java/org/zanata/model/HAccount.java#L115)):*
> ```Java
>     // @UserEnabled
> 
>     @Field
> ```

I believe this will create a hibernate search index field. What we need here is just a database index.  I think what you need is @javax.persistence.Index

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/504)*

huangp:
Reviewed 2 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/504)*

seanf:
@aeng @huangp I don't think mysql is smart enough to use the two separate indexes the way you're assuming. I think it will only use one index for any given query, and then trawl through all the results checking the other column value. 

You probably want to index `(enabled, username)` in that order. This should also let you find all enabled users quickly, or all enabled users with a common prefix. I recommend analysing some SQL queries to ensure that the query analyser uses the index(es) as expected.

aeng:
@huangp can you review again? did another commit to index both fields.

seanf:
Reviewed 1 of 1 files at r2.

---

*[server/zanata-model/src/main/java/org/zanata/model/HAccount.java, line 115 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/504#-KskKq0dBpMLRZpwjOsZ:-Kspef5OtYcI1OgbR5qb:b-y8gd09) ([raw file](https://github.com/zanata/zanata-platform/blob/55190479cc8e6bef30988d50af3c4474d0b8a715/server/zanata-model/src/main/java/org/zanata/model/HAccount.java#L115)):*

I believe this will create a hibernate search index field. What we need here is just a database index.  I think what you need is @javax.persistence.Index
</blockquote></details>

@aeng 
It doesn't look like this Hibernate Search field is used in any queries. 

If we need to exclude disabled users from *all* Hibernate Search results, we probably need another interceptor similar to `org.zanata.hibernate.search.TextFlowTargetEntityIndexingInterceptor` which SKIPs obsolete entries. *However*, we may want to distinguish disabled users from soft-deleted users, in which case we would need a second boolean field (eg `obsolete`) for soft deletions.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/504)*

aeng:

---

*[server/zanata-model/src/main/java/org/zanata/model/HAccount.java, line 115 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/504#-KskKq0dBpMLRZpwjOsZ:-KssVk84V1nhOyNJNFBS:b2ke7um) ([raw file](https://github.com/zanata/zanata-platform/blob/55190479cc8e6bef30988d50af3c4474d0b8a715/server/zanata-model/src/main/java/org/zanata/model/HAccount.java#L115)):*

@aeng 
It doesn't look like this Hibernate Search field is used in any queries. 

If we need to exclude disabled users from *all* Hibernate Search results, we probably need another interceptor similar to `org.zanata.hibernate.search.TextFlowTargetEntityIndexingInterceptor` which SKIPs obsolete entries. *However*, we may want to distinguish disabled users from soft-deleted users, in which case we would need a second boolean field (eg `obsolete`) for soft deletions.
</blockquote></details>

Removing this annotation since we don't really need it. TextFlowTargetEntityIndexingInterceptor is for lucene index which HAccount does not required. The db index should be enough for increasing the performance?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/504)*

seanf:
Ah, in that case... We currently have the Hibernate Search annotation `@Indexed` on these classes:
* **HAccount** - queried by nothing that I can find - commented out it AccountDAO
* **HGlossaryEntry** - queried by nothing that I can find
* HGlossaryTerm - queried by GlossaryDAO
* HProject - queried by ProjectDAO
* **HProjectIteration** - queried by nothing that I can find
* HTextFlowTarget - queried by TranslationMemoryServiceImpl
* TransMemoryUnit - queried by TranslationMemoryServiceImpl

If we're not planning to use (or maintain) the Lucene-related code in these classes, we should remove **all** their Hibernate Search annotations. Leaving them half-supported makes the code harder to understand (eg why is this `@Indexed` but it ignores deletions?) and also causes unnecessary indexing at runtime.

Even if we're not using Hibernate Search for HAccount, we still need to think about whether we need that extra "obsolete" boolean. Can we really say that the current "disabled" flag is suitable for soft-deleting accounts? Because that is how it is sometimes used. I think they're already two different concepts.

aeng:
Since this pull request is mainly for fixing HAccount, I will not touch other files for the time being.
For soft delete, I think currently the "disabled" flag is sufficient enough until the time where we need to delete an account which I prefer hard delete with strong warning. But we can discuss that when it is needed.

huangp:
Review status: 2 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[server/zanata-model/src/main/java/org/zanata/model/HAccount.java, line 56 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/504#-Ksui1gxrta9SdzUlp19:-Ksui1gxrta9SdzUlp1A:b-if9oil) ([raw file](https://github.com/zanata/zanata-platform/blob/55190479cc8e6bef30988d50af3c4474d0b8a715/server/zanata-model/src/main/java/org/zanata/model/HAccount.java#L56)):*
> ```Java
> @Cacheable
> @Table(uniqueConstraints = @UniqueConstraint(columnNames = "username"))
> @Indexed
> ```

username field is still annotated with ```@Field```. Let's keep this for the time being as this PR is about to add the database index.
You need to add ```@javax.persistence.Index(name = "Idx_enabled_username", columnList = "username, enabled") to match what's in liquibase.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/504)*

seanf:
Reviewed 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[server/zanata-model/src/main/java/org/zanata/model/HAccount.java, line 56 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/504#-Ksui1gxrta9SdzUlp19:-Ksuj-BMpfB1x6ofydoY:b-3aod9k) ([raw file](https://github.com/zanata/zanata-platform/blob/55190479cc8e6bef30988d50af3c4474d0b8a715/server/zanata-model/src/main/java/org/zanata/model/HAccount.java#L56)):*

username field is still annotated with ```@Field```. Let's keep this for the time being as this PR is about to add the database index.
You need to add ```@javax.persistence.Index(name = "Idx_enabled_username", columnList = "username, enabled") to match what's in liquibase.
</blockquote></details>

To me, this PR is not just about adding the index. 

It's really about fixing the problems with https://github.com/zanata/zanata-platform/pull/485, problems which are about not handling disabled accounts consistently. Are we planning to filter out disabled accounts or not? Different parts of our codebase answer that question differently, because our database queries and Lucene indexing don't match. If we get rid of the Lucene indexing, that problem goes away.

---

*[server/zanata-model/src/main/java/org/zanata/model/HAccount.java, line 115 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/504#-KskKq0dBpMLRZpwjOsZ:-Ksuju0KnOYmYl6onVuq:b-mwl3j1) ([raw file](https://github.com/zanata/zanata-platform/blob/55190479cc8e6bef30988d50af3c4474d0b8a715/server/zanata-model/src/main/java/org/zanata/model/HAccount.java#L115)):*
<details><summary><i>Previously, aeng (Alex Eng) wrote…</i></summary><blockquote>

Removing this annotation since we don't really need it. TextFlowTargetEntityIndexingInterceptor is for lucene index which HAccount does not required. The db index should be enough for increasing the performance?
</blockquote></details>

When I mentioned an IndexingInterceptor, I was under the impression we used this Lucene indexing for some HAccount queries. As we don't, it's better just to remove it.

---

*[server/zanata-model/src/main/java/org/zanata/model/HAccount.java, line 45 at r3](https://reviewable.io:443/reviews/zanata/zanata-platform/504#-KsukbIaJXGu7zShaJcf:-KsukbIaJXGu7zShaJcg:b-to4jzh) ([raw file](https://github.com/zanata/zanata-platform/blob/4e528f8519028fd80af10adb55645ca8f8d3e89a/server/zanata-model/src/main/java/org/zanata/model/HAccount.java#L45)):*
> ```Java
> import org.hibernate.annotations.CacheConcurrencyStrategy;
> import org.hibernate.annotations.NaturalId;
> import org.hibernate.search.annotations.Field;
> ```

This annotation should probably be removed too.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/504)*

seanf:
I have created https://zanata.atlassian.net/browse/ZNTA-2197 for the other dead code (obsolete Hibernate Search annotations).

aeng:
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[server/zanata-model/src/main/java/org/zanata/model/HAccount.java, line 45 at r3](https://reviewable.io:443/reviews/zanata/zanata-platform/504#-KsukbIaJXGu7zShaJcf:-KtE4B1GL8GZHi7IjCQq:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/4e528f8519028fd80af10adb55645ca8f8d3e89a/server/zanata-model/src/main/java/org/zanata/model/HAccount.java#L45)):*

This annotation should probably be removed too.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/504)*

seanf:
:lgtm:

---

Reviewed 2 of 2 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/504#-:-KtERUysIMsEGo_d9hgF:bnfp4nl)*

spathare21:
Verified this with the help of PR #485, tested profile page of an enabled user with login and without login. let me know if need to verify anything else.

PR #505, opened 2017-08-30
fix(Jenkinsfile): RejectedAccessException

seanf:
Fixes error when build fails:

org.jenkinsci.plugins.scriptsecurity.sandbox.RejectedAccessException: Scripts not permitted to use method groovy.lang.GroovyObject invokeMethod java.lang.String java.lang.Object (org.zanata.jenkins.Notifier error hudson.AbortException)

---
This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-platform/505)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -316,7 +316,7 @@ timestamps {
         sh "git clean -fdx"
       } catch (e) {
         echo("Caught exception: " + e)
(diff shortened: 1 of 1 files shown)


seanf:
Note: build only failed due to cancellation (to test the change in error handling).

PR #198, opened 2017-02-15
chore: run Zanata and DB via docker-compose

seanf:
JIRA issue URL: *paste URL here*

*[Add a short description of what is changed here.]*

## Reviewer tasks

The following is based on the
[Zanata Development Guidelines](https://github.com/zanata/zanata-platform/wiki/Development-Guidelines).
As a reviewer, you should check the guidelines regularly for updates, and just
to refresh your memory.

The reviewer can use this list for reference to make sure
they have considered all the important points.

Use the checkboxes or not, whatever works best for you.

 - Code quality
   - [ ] Code passes style check (checkstyle/eslint)
   - [ ] Code is clear and concise
   - [ ] UI code is internationalised
   - [ ] Code has adequate comments where appropriate
   - Code is in an appropriate place
     - [ ] Classes are in appropriate packages
     - [ ] Code fits with class's responsibilities (SRP)
   - [ ] Configuration files are documented enough
   - If code is removed
     - [ ] No remaining code references the removed code
       - [ ] No orphaned rewrite rules
       - [ ] No orphaned config settings
     - [ ] No remaining comments refer to the removed code
     - [ ] Unused imports and libraries are removed
 - Testing
   - [ ] New code has tests
   - [ ] Changed code has tests
   - [ ] All tests pass
   - [ ] Test coverage stable or increased
 - Documentation
   - [ ] New features are documented
   - [ ] Docs removed for deleted features
   - [ ] Docs updated for all changed features
   - [ ] Developer/sysadmin docs updated if appropriate
 - If there are new dependencies
   - [ ] Does each new dependency pass all the
         [Considerations for new dependencies](https://github.com/zanata/zanata-platform/wiki/Development-Guidelines#new-technologies-and-dependencies)?

----
*This template can be updated in .github/PULL_REQUEST_TEMPLATE.md*

---
This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-platform/198)

PR #548, opened 2017-10-03
chore: run Zanata and DB via docker-compose

seanf:
Revival of https://github.com/zanata/zanata-platform/pull/198

Note to QA: this change introduces `rundev2.sh` which works much the same as `rundev.sh` but with fewer options. It also starts and stops the database automatically.

---
This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-platform/548)


seanf:
@carlosmunoz This is a rebase of #198 (which you reviewed) with just one new commit, 4eb5a60.

carlosmunoz:
Still looks good. Perhaps the only thing to note is that there is no guarantee that the database will be up before the Jboss server attempts to connect to it. In practice it should be ready as the jboss server is not the quickest to start.

---

Reviewed 5 of 7 files at r3, 1 of 1 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/548#-:-KvXYcV4e7DaaivZ82XV:bazrbv6)*

seanf:
@carlosmunoz No_QA?

seanf:
@carlosmunoz Another review please?

Re: the database startup problem, I'm suggesting we ignore that problem for now, and uncomment the `healthcheck` in `docker-compose.yml` when Fedora 27 is out (because F26 and F27's docker-compose will support `healthcheck`, but F25 doesn't).

carlosmunoz:
Reviewed 1 of 7 files at r3, 2 of 2 files at r5, 1 of 1 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/548)*

carlosmunoz:
Reviewed. 
Um... why not just enable it and make it clear this will only work with a specific version of docker-compose. If you are on Fedora 25, then you either need to upgrade, or manually install a latest version of docker compose. Lets just enable it and benefit from it starting now.

seanf:
@carlosmunoz Thanks.

For one thing, _I'm_ still on Fedora 25. (And I'm not sure it's easy to install docker.io's docker-compose without breaking Fedora's docker installation.)

carlosmunoz:
I went with docker.io's one when I was on F25. You definitely don't want to try and set them both up, but then again, why would you want both versions, if you have the most recent one? Stop being so cautious and upgrade to F26 already :stuck_out_tongue:

PR #565, opened 2017-10-09
Log all Weld exceptions in Docker runs

seanf:
[LOGMGR-45](https://issues.jboss.org/browse/LOGMGR-45) implies that we should now be able to inherit the filters as we were trying to do, but it's unclear if it can be done in the appserver configuration: https://developer.jboss.org/message/978927#978927

This configuration change adds the filter to each of the log categories which Weld might use for logging exceptions as DEBUG.

This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-platform/565)


huangp:
:lgtm: +(No QA \(2nd vote\)) 

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/565#-:-L0NDXMs0rEr18d6WZ9L:b69w6nj)*

PR #598, opened 2017-10-25
feat(glossary): JSON parser for glossary

djansen-redhat:
JIRA issue URL: n/a

Allow the upload of a glossary file in a json format.

## Checklist

- JIRA link
- Check target branch

----


server/services/src/main/java/org/zanata/service/impl/GlossaryFileServiceImpl.java
```diff
@@ -116,8 +119,40 @@ public GlossaryProcessed saveOrUpdateGlossary(
...
+                boolean success = false;
+                Optional<GlossaryTerm> srcTerm = getSourceTerm(entry);
+                if (!srcTerm.isPresent()) {
+                    continue;
+                }
+                // Assume first import, give it a new ID based on term and pos
+                String newId = srcTerm.get().getContent().concat("-").concat(entry.getPos());
+
+                entry.setExternalId(newId);
+                if (entryExists(entry)) {
```
djansen-redhat: @aeng @seanf I'm sure I'm doing something evil here. Opinions?

aeng: What is that id for? would be better to use hash on those fields?

djansen-redhat: It's for external management, so terms can be uploaded with an id for tracking and updating. e.g.
```
{ terms:[{
  id: "thing-noun",
  term: "thing",
  description: "a whatsit"
}] }
```
This allows for instances where you have multiple definitions, like "bat" having multiple nouns and verbs, and so that if you fix a typo in a description you don't end up with duplicate entries.
(This was a side project of mine, so I should make a jira that properly explains it.)

common/zanata-adapter-glossary/src/main/java/org/zanata/adapter/glossary/GlossaryJsonReader.java
```diff
@@ -0,0 +1,158 @@
...
+
+/**
+ * @author Damian Jansen <a href="mailto:djansen@redhat.com">djansen@redhat.com</a>
+ **/
+public class GlossaryJsonReader {
+    private final LocaleId srcLang;
+
+    private final static String TERM = "term";
+    private final static String TRANSLATIONS = "translations";
+    private final static String[] POSSYNONYMS = {"pos", "partofspeech", "part of speech"};
```
djansen-redhat: Is this convenience greatly unnecessary?

server/services/src/main/java/org/zanata/service/impl/GlossaryFileServiceImpl.java
aeng: In texflow, we use hash as the external resid. I think we should do the same here. Hashing of the content and pos

common/zanata-adapter-glossary/src/main/java/org/zanata/adapter/glossary/GlossaryJsonReader.java
aeng: Not really if we want to make it "user friendly". But we will need to clearly state what the system recognised as the part of speech title.

server/services/src/main/java/org/zanata/service/impl/GlossaryFileServiceImpl.java
djansen-redhat: The problem is you can have e.g. multiple "bat" terms that are nouns, which will incorrectly resolve to the same hash. Further, you can change the term's POS or Description in the glossary editor and reimport the same glossary file, resulting in duplication - or do this externally with the same result.

djansen-redhat: Would it be more sensible to get a list of every id that matches "term-pos*" and match + increment locally?

aeng: Yes and that has always been the problem with our glossary. That is why we use the db:id instead. The hash column is used for checking when new glossary entry is being created (on server side).

PR #635, opened 2017-12-15
WIP: convert container and component editor css to less - remove suit…

kathryngo:
Attempts to convert root/index.css in editor to less (and resolving all linting issues). Breaks and looks like the following with a clean build:

brokencss

None of the css is generating. I cannot work out why this is happening. Would someone mind looking at my webpack and css set up to see what is going on? Thanks.

I have fixed the suitcss issues so it is not that causing the problems. No console feedback either, but the css/les files are definitely not being loaded. Please note, I could get my changes to work until I restart my machine and run a clean build 

*[description]*

## Checklist

- JIRA link
- Check target branch

----

---
This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-platform/635)


kathryngo:
Unfortunately the editor css must stay as css and not .less files because LESS + relative URLS + css-loader is incompatible (see here: https://github.com/webpack-contrib/less-loader/issues/109#issuecomment-25379733)

Must be why editor was done purely in css to begin with.

Can slowly migrate the editor code to the frontend styles.less by using a wrappers like this:

`#editor { //css here }`

but I think they should be separate to reduce page load times. The less files I tried to create could not use the variables.less file so the inline variables in the editor css files will have to suffice for the moment. Any less css from editor will not generate - they are replaced by style.less from frontend.

I think as a practice the editor css and frontend css should be seperate until such times that we are using the same components across frontend and editor.

TLDR: Keep frontend and editor css directories. Isolate css to component level where possible - about optimisation above everything else. Any attempt to turn editor css to less results in missing css.

PR #685, opened 2018-01-23
test(unittest): ActivateAction test

djansen-redhat:
Unit test for ActivateAction, except for the activate function.
Changes ActivateAction from a field inject to a constructor inject

JIRA issue URL: n/a

## Checklist

- JIRA link
- Check target branch

----

---
This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-platform/685)

### DIFF
--- a/server/services/src/main/java/org/zanata/action/ActivateAction.java
+++ b/server/services/src/main/java/org/zanata/action/ActivateAction.java
@@ -48,20 +48,33 @@ public class ActivateAction implements Serializable {
 
     private static final long serialVersionUID = -8079131168179421345L;
     private static final int LINK_ACTIVE_DAYS = 1;
-    @Inject
+
     private GroupedConversation conversation;
-    @Inject
     private AccountActivationKeyDAO accountActivationKeyDAO;
-    @Inject
     private IdentityManager identityManager;
-    @Inject
     private UrlUtil urlUtil;
-    @Inject
+
     private FacesMessages facesMessages;
     private String activationKey;
     private HAccountActivationKey key;
     private String resetPasswordKey;
     // @Begin(join = true)
+
+    @Inject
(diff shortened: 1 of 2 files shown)


seanf:
Reviewed 2 of 2 files at r1.

---

*[server/services/src/test/java/org/zanata/action/ActivateActionTest.java, line 79 at r1](https://reviewable.io:443/reviews/zanata/zanata-platform/685#-L3WWY1rqO-mewmtrrJN:-L3WWY1rqO-mewmtrrJO:bdiyav5) ([raw file](https://github.com/zanata/zanata-platform/blob/99853171edd31e50fd51174b8c7952cb52240893/server/services/src/test/java/org/zanata/action/ActivateActionTest.java#L79)):*
> ```Java
>     }
> 
>     @Test(expected = KeyNotFoundException.class)
> ```

I never recommend using the `expected` attribute on `@Test` (except for one line tests, maybe). It doesn't care when the exception happens, so the exception might not mean what you think it does.

try-catch inside the test is more accurate, even if it is a bit more verbose. We should probably set up a little helper method which takes a lambda and an exception class, if someone hasn't written it already.

```java
void expectThrowable(Class<? extends Throwable t, Runnable r) {
    try {
        r.run();
        Assert.fail("didn't get expected Throwable "+t);
    } catch (/* exercise for the reader */) {
        // ignore
    }
}
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/685)*

djansen-redhat:
@seanf this one was updated to use try-catch

seanf:
:lgtm:

---

Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/685#-:-L5SDA6HrcltyddDyYxp:bnfp4nl)*

PR #694, opened 2018-01-29
feat(ZNTA-2349): Add option for Hibernate Search to use infinispan-directory-provider instead of file system

oytuntez:
JIRA issue URL: https://zanata.atlassian.net/browse/ZNTA-2349

Allows using Infinispan as the directory provider for Hibernate Search. This also makes Infinispan the default directory provider.

* PS: Our previous PR merges from your upstream is also in the history of this PR. I couldn't rebase the master, I hope this works.

## Checklist

- JIRA link
- Check target branch

----

---
This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-platform/694)


seanf:
Thanks @oytuntez!

Last time we tried this, I think we found that the performance of the Infinispan directory was a lot slower (for non-clustered use cases), so the plan was to make this an **option** (so that non-clustered users could stick to the file system). [Thus requiring changes in the documentation and perhaps a system property to choose the option.] But our attempt was a while ago, so before we go down that path: have you got any idea of the performance impact of this change?

Regarding all those merges: We should be able to use squash-and-merge in this case, but _why_ do you need all those merges? They're just going to make the history even harder to read (especially if the use of Backstroke is going to keep creating more and more in future).

The repo https://github.com/MotaWord/zanata-platform doesn't appear to include **anything** except upstream changes (from https://github.com/zanata/zanata-platform) plus merge nodes [and this one change]. If that repo doesn't have any commits of its own, it should be able to use fast-forward merges. I don't really understand what value that repo is giving you. Does it have other, non-public branches or something?

The build failed because of maven enforcer rules:

```
20:49:34 [WARNING] Rule 6: org.apache.maven.plugins.enforcer.RequireUpperBoundDeps failed with message:
20:49:34 Failed while enforcing RequireUpperBoundDeps. The error(s) are [
20:49:34 Require upper bound dependencies error for org.infinispan:infinispan-core:5.1.6.FINAL paths to dependency are:
20:49:34 +-org.zanata:functional-test:4.5.0-SNAPSHOT
20:49:34   +-org.zanata:services:4.5.0-SNAPSHOT
20:49:34     +-org.infinispan:infinispan-directory-provider:8.1.0.Final
20:49:34       +-org.infinispan:infinispan-core:5.1.6.FINAL
20:49:34 and
20:49:34 +-org.zanata:functional-test:4.5.0-SNAPSHOT
20:49:34   +-org.zanata:services:4.5.0-SNAPSHOT
20:49:34     +-org.infinispan:infinispan-directory-provider:8.1.0.Final
20:49:34       +-org.infinispan:infinispan-lucene-directory:8.1.0.Final
20:49:34         +-org.infinispan:infinispan-core:5.1.6.FINAL (managed) <-- org.infinispan:infinispan-core:8.1.0.Final
```

I would try to fix them, but I don't want to mess with your `master` branch. I always recommend using a separate branch for pull requests, other than `master`.

server/pom.xml
```diff
@@ -1336,7 +1336,7 @@
         <groupId>org.infinispan</groupId>
         <artifactId>infinispan-directory-provider</artifactId>
         <version>${infinispan.version}</version>
-        <scope>provided</scope>
+        <scope>compile</scope>
```
seanf: This would be better as `${hibernate.search.scope}`

seanf:
@oytuntez I think this will fix the enforcer error. (Tested with `mvn enforcer:enforce -Denforcer.skip=false -DstaticAnalysis`.)
```diff
diff --git a/server/pom.xml b/server/pom.xml
index 9f1bf2bcdc..9673dad064 100644
--- a/server/pom.xml
+++ b/server/pom.xml
@@ -1336,7 +1336,14 @@
         <groupId>org.infinispan</groupId>
         <artifactId>infinispan-directory-provider</artifactId>
         <version>${infinispan.version}</version>
-        <scope>compile</scope>
+        <scope>${hibernate.search.scope}</scope>
+      </dependency>
+
+      <dependency>
+        <groupId>org.infinispan</groupId>
+        <artifactId>infinispan-core</artifactId>
+        <version>${infinispan.version}</version>
+        <scope>provided</scope>
       </dependency>
 
       <dependency>


```

oytuntez:
Hey Sean,

Thanks for the guidance.

I want to cancel (or suspend) this PR for now, because even though this worked perfectly with a single instance, we weren't able to get the Infinispan cache up and running in a clustered way for Hibernate Search.

Yesterday I started upgrading Hibernate ORM, Search and Validator to their latest versions. There are some changes, but not too many. If this works out, we'll use Elasticsearch for distributed Search backend. I believe this solves the problem in a more sustainable way (rather than 1) dealing with Infinispan configs, distribution and clustering, which can be cumbersome, 2) we get to upgrade our data modules.

How does this sound?

Regarding our Zanata fork  https://github.com/MotaWord/zanata-platform: because we change things that wouldn't be related to Zanata's target group, we keep this repo for PR purposes actually. Locally, we only cherry-pick from our custom repo (which is hosted at BitBucket). Trying to maintain the same fork in the same repo could work, but there is a lot of overhead doing it especially if we want to break things fast and stupid without affecting upstream :) I hope this makes sense and of course, we are open to suggestions.

Oytun

PR #702, opened 2018-02-01
fix(ZNTA-2371) fix dropdowns on Project Version pages

kathryngo:
JIRA issue URL: https://zanata.atlassian.net/browse/ZNTA-2371

Fix **project version page dropdowns and project page dropdowns** (by referring to generated id name) so that the dropdowns show all content and are not cut off at the bottom of the container. 

Also fixed the **group languages selection dropdown** which had the same issue.

group-settings

This is all due to the height of the tab-panels on these pages being restricted to its content so that when the dropdown is visible, it is cut off from view.

## Checklist

- JIRA link
- Check target branch

----

---
This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-platform/702)


server/zanata-frontend/src/app/styles/legacy.less
```diff
@@ -134,4 +134,11 @@
     word-wrap: break-word;
     white-space: pre-wrap;
   }
+ //fixes to Versions page
+ form#j_idt38.l--push-bottom-0 ~ .g.u-fullWidth {
+     height: 100vh;
+  }
+  #versions_form ~ ul.dropdown__content li {
```
seanf: I think this should be outdented.

kathryngo: It needs to be indented correctly (which I didn't do here, just fixed it) so that is is part of the wrapper `.new-zanata`. This ensures everything in this legacy file applies only to the jsf pages which have that top-level class.

And I will make sure to run the functional tests before merge.

seanf:
:lgtm:

---

Reviewed 2 of 3 files at r1, 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/702#-:-L4EmlXTOk3nYd9UEjgi:bnfp4nl)*

spathare21:
Verified merge once require status for codecov pass

spathare21:
@kathryngo is it possible to fix the project pages dropdown in the same PR or need to file new issue 

screenshot from 2018-02-01 11-26-40

screenshot from 2018-02-01 11-27-15

kathryngo:
@spathare21 please file a new issue

spathare21:
ok sure thanks @kathryngo

kathryngo:
@spathare21 I think I might have to fix it in this PR because of failing tests.

seanf:
Reviewed 1 of 3 files at r1, 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[server/zanata-frontend/src/app/styles/legacy.less, line 138 at r3](https://reviewable.io:443/reviews/zanata/zanata-platform/702#-L4IwFLChWfepUU9CK-4:-L4IwFLChWfepUU9CK-5:b-l3scpv) ([raw file](https://github.com/zanata/zanata-platform/blob/4cbe561333a19182216e88f7dca4e987f2572500/server/zanata-frontend/src/app/styles/legacy.less#L138)):*
> ```Less
>   }
>   //fixes to Versions/Projects page
>   form#j_idt38.l--push-bottom-0 ~ .g.u-fullWidth,
> ```

is the ID `j_idt38` auto-generated? It looks fragile. Perhaps we need to assign a proper ID to the relevant `form` element(s) in XHTML.

---

*[server/zanata-frontend/src/app/styles/legacy.less, line 139 at r3](https://reviewable.io:443/reviews/zanata/zanata-platform/702#-L4IxD6wN0mp5J5Y2sl6:-L4IxD6wN0mp5J5Y2sl7:b-q1v6ua) ([raw file](https://github.com/zanata/zanata-platform/blob/4cbe561333a19182216e88f7dca4e987f2572500/server/zanata-frontend/src/app/styles/legacy.less#L139)):*
> ```Less
>   //fixes to Versions/Projects page
>   form#j_idt38.l--push-bottom-0 ~ .g.u-fullWidth,
>   li#projects.is-active  {
> ```

This seems like another overly-generic selector, since it will target any `li` with the id `projects.is-active`, regardless of the page. That if I'm reading my CSS correctly. Then again, this is Less, so I could definitely be missing something.

---

*[server/zanata-frontend/src/app/styles/legacy.less, line 143 at r3](https://reviewable.io:443/reviews/zanata/zanata-platform/702#-L4IwePo7yiXzd1euniA:-L4IwePo7yiXzd1euniB:b-m35pqv) ([raw file](https://github.com/zanata/zanata-platform/blob/4cbe561333a19182216e88f7dca4e987f2572500/server/zanata-frontend/src/app/styles/legacy.less#L143)):*
> ```Less
>   }
>   // fix to settings dropdowns on languages tab
>   li#settings ul.tabsContent {
> ```

This seems to be a pretty generic selector for such a specific comment. This seems like another case which deserves XHTML changes. Maybe we should have a class name which mentions "languages"?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/702)*

kathryngo:
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[server/zanata-frontend/src/app/styles/legacy.less, line 138 at r3](https://reviewable.io:443/reviews/zanata/zanata-platform/702#-L4IwFLChWfepUU9CK-4:-L4J-9C8ChPSJ-hTmaAn:b-anrq9n) ([raw file](https://github.com/zanata/zanata-platform/blob/4cbe561333a19182216e88f7dca4e987f2572500/server/zanata-frontend/src/app/styles/legacy.less#L138)):*

is the ID `j_idt38` auto-generated? It looks fragile. Perhaps we need to assign a proper ID to the relevant `form` element(s) in XHTML.
</blockquote></details>

Yep, fixing these pages are tricky. The container that needs to be targeted is is not a parent, and is generated through some higher level component that is used across a heap of pages.

---

*[server/zanata-frontend/src/app/styles/legacy.less, line 139 at r3](https://reviewable.io:443/reviews/zanata/zanata-platform/702#-L4IxD6wN0mp5J5Y2sl6:-L4J-RWNq3YyPdpVaC90:b-17xrmi) ([raw file](https://github.com/zanata/zanata-platform/blob/4cbe561333a19182216e88f7dca4e987f2572500/server/zanata-frontend/src/app/styles/legacy.less#L139)):*

This seems like another overly-generic selector, since it will target any `li` with the id `projects.is-active`, regardless of the page. That if I'm reading my CSS correctly. Then again, this is Less, so I could definitely be missing something.
</blockquote></details>

I have looked and this is the only place it is being used. Once again it is the issue with not having the ability to add classnames to the necessary div because of the autogenerated code

---

*[server/zanata-frontend/src/app/styles/legacy.less, line 143 at r3](https://reviewable.io:443/reviews/zanata/zanata-platform/702#-L4IwePo7yiXzd1euniA:-L4J-cZBTsZX6MM8yRHx:bx7a50k) ([raw file](https://github.com/zanata/zanata-platform/blob/4cbe561333a19182216e88f7dca4e987f2572500/server/zanata-frontend/src/app/styles/legacy.less#L143)):*

This seems to be a pretty generic selector for such a specific comment. This seems like another case which deserves XHTML changes. Maybe we should have a class name which mentions "languages"?
</blockquote></details>

Would do if I could get it to target the correct level of code - these pages only cover what is in the containers, not the container itself. 

Looks like the IDs are not used anywhere else, made sure to check.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/702)*

seanf:
We should talk about better options for these CSS/JSF issues in the next sprint retrospective, but otherwise LGTM.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/702#-:-L4J6lIQ39klokfzd9SV:b-2pgfg7)*

aeng:
@kathryngo @seanf we need to use id or define a class to replace `j_idt38` as it is generated and might change overtime.

kathryngo:
@seanf @aeng I wish it was that simple. For the versions panel I need to target the level with the class `g u-fullWidth` which is higher than I can access via the page for the versions tab (see image).
The top level is `panel` in the versions file. So **we would need some ID at the level that the `g` class is populated** and that is the problem.

abs

It is not possible in css to select the previous element, only the next element (using ~) and I cannot use any selectors to target them from parent divs since there is nothing unique to this page. 
Applying any changes here that affect the other pages causes the build to fail.

seanf:
@kathryngo I haven't tested this, but what if we did something like this (I'm pretty sure this is the right file to change):

```patch
diff --git a/server/zanata-war/src/main/webapp/project/project.xhtml b/server/zanata-war/src/main/webapp/project/project.xhtml
index d4ed0333c1..f2d2e0949b 100644
--- a/server/zanata-war/src/main/webapp/project/project.xhtml
+++ b/server/zanata-war/src/main/webapp/project/project.xhtml
@@ -52,11 +52,12 @@
 
   <ui:fragment
     rendered="#{identity.hasPermission(projectHome.instance, 'update')}">
-    <zanata:confirm-modal id="exportTMXVersionModal"
-      title="#{msgs['jsf.iteration.ExportTMXIter']}"
-      message="#{msgs['jsf.ConfirmExportTMXIter']}"
-      onresponse="onExportTMXVersionResponse" />
-
+    <div id="exportTMXVersionModalDiv">
+      <zanata:confirm-modal id="exportTMXVersionModal"
+        title="#{msgs['jsf.iteration.ExportTMXIter']}"
+        message="#{msgs['jsf.ConfirmExportTMXIter']}"
+        onresponse="onExportTMXVersionResponse" />
+    </div>
     <script type="application/javascript">
       var selectedVersion;
       function setSelectedVersion(version) {


```

seanf:
Correction, that should be:

```patch
diff --git a/server/zanata-war/src/main/webapp/project/project.xhtml b/server/zanata-war/src/main/webapp/project/project.xhtml
index d4ed0333c1..234a7bf014 100644
--- a/server/zanata-war/src/main/webapp/project/project.xhtml
+++ b/server/zanata-war/src/main/webapp/project/project.xhtml
@@ -50,13 +50,13 @@
 
 </script>
 
+<div id="exportTMXVersionModalDiv">
   <ui:fragment
     rendered="#{identity.hasPermission(projectHome.instance, 'update')}">
     <zanata:confirm-modal id="exportTMXVersionModal"
       title="#{msgs['jsf.iteration.ExportTMXIter']}"
       message="#{msgs['jsf.ConfirmExportTMXIter']}"
       onresponse="onExportTMXVersionResponse" />
-
     <script type="application/javascript">
       var selectedVersion;
       function setSelectedVersion(version) {
@@ -69,8 +69,6 @@
         zanata.modal.hide('#exportTMXVersionModal');
       }
     </script>
-
-
   </ui:fragment>
 
   <h:form styleClass="l--push-bottom-0">
@@ -89,8 +87,8 @@
     <a4j:jsFunction name="resetProjectMembers"
       render="people-size,people_form" limitRender="true"
       action="#{projectHomeAction.clearCachedMembershipData}"/>
-
   </h:form>
+</div>
 
 <div class="g u-fullWidth">
   <div class="g__item w--1-m w--3-8-l w--1-3 l--push-bottom-half">

```

seanf:
At this point the choice of id `exportTMXVersionModalDiv` is looking a bit silly, but it shows the general idea.

aeng:
:lgtm:

---

Reviewed 1 of 4 files at r1, 2 of 3 files at r5.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/zanata/zanata-platform/702#-:-L4cai4D-jjQnEbWV1bx:bnfp4nl)*

PR #709, opened 2018-02-02
Type check JS by default

seanf:
Configure TypeScript to check JS files too. I have fixed type errors in some files, and disabled checking in the other 171 or so.

---
This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-platform/709)


aeng:
:lgtm:  Ideally we should remove all the `ts-nocheck`. But that ok for now.

---

Reviewed 223 of 223 files at r1, 6 of 6 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/709#-:-L5Qu_w-ddrjWzxsiDax:b-5t5eb2)*

PR #715, opened 2018-02-06
fix: storybook frontend and editor

aeng:
https://zanata.atlassian.net/browse/ZNTA-2388

*[description]*

## Checklist

- JIRA link
- Check target branch

----

---
This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-platform/715)


efloden:
:lgtm:

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/715#-:-L4d_l84A3EW6VEb6XnW:bnfp4nl)*

efloden:
storybook-frontend appears to be working now, but the storybook-editor has an error:
 ReferenceError: exports is not defined
storybookerror

seanf:
@aeng I've added my `var exports = {}` change to the branch. Together, I think the changes do work around the storybook-editor problem; in fact there no longer seems to be a need to use lodash.keys or to re-add the babel dependencies - the yarn warnings about babel-core and babel-runtime don't seem to be significant.

So the question is why using JSON.stringify in app/editor/api/index.js causes the `api` module to come back as empty when running via storybook. Eliminating JSON.stringify and `var exports = {}` both seem like kludgey workarounds which could break again any time.

aeng:
@seanf Might be related to https://cdroulers.com/blog/2015/04/22/typescript-properties-and-json-stringify
I think we should be fine with the fix in this branch for now.

efloden:
Reviewed 6 of 9 files at r2, 1 of 1 files at r3, 2 of 3 files at r4.

---

*[server/zanata-frontend/src/app/editor/api/index.js, line 120 at r4](https://reviewable.io:443/reviews/zanata/zanata-platform/715#-L4mbOk1S7D3LNOY525s:-L4mbOk24fTbF85M-ePY:b-mrip5k) ([raw file](https://github.com/zanata/zanata-platform/blob/15055afe5743975f8431f183e31e5069ed805902/server/zanata-frontend/src/app/editor/api/index.js#L120)):*
> ```JavaScript
>     },
>     mode: 'cors',
>     body: JSON.stringify({
> ```

There are many more instances of JSON.stringify in the action files, should we try to replace all of them with stableStringify?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/715)*

efloden:
:lgtm:

---

---

*[server/zanata-frontend/src/app/editor/api/index.js, line 120 at r4](https://reviewable.io:443/reviews/zanata/zanata-platform/715#-L4mbOk1S7D3LNOY525s:-L4mcC_j6Kzc8B-Vvfd3:b32w0x) ([raw file](https://github.com/zanata/zanata-platform/blob/15055afe5743975f8431f183e31e5069ed805902/server/zanata-frontend/src/app/editor/api/index.js#L120)):*
<details><summary><i>Previously, efloden (Earl Floden) wrote…</i></summary><blockquote>

There are many more instances of JSON.stringify in the action files, should we try to replace all of them with stableStringify?
</blockquote></details>

OK

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/715#-:-L4mcFr-eTMqMxTR0z1K:bnfp4nl)*

aeng:

---

*[server/zanata-frontend/src/app/editor/api/index.js, line 120 at r4](https://reviewable.io:443/reviews/zanata/zanata-platform/715#-L4mbOk1S7D3LNOY525s:-L4mcWYzrRj8r6XyxEOW:b-stdatb) ([raw file](https://github.com/zanata/zanata-platform/blob/15055afe5743975f8431f183e31e5069ed805902/server/zanata-frontend/src/app/editor/api/index.js#L120)):*
<details><summary><i>Previously, efloden (Earl Floden) wrote…</i></summary><blockquote>

OK
</blockquote></details>

I can only see one - which is the one I replaced

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/715)*

aeng:

---

*[server/zanata-frontend/src/app/editor/api/index.js, line 120 at r4](https://reviewable.io:443/reviews/zanata/zanata-platform/715#-L4mbOk1S7D3LNOY525s:-L4mfPbhukH7D0IBMXEE:bko27s7) ([raw file](https://github.com/zanata/zanata-platform/blob/15055afe5743975f8431f183e31e5069ed805902/server/zanata-frontend/src/app/editor/api/index.js#L120)):*
<details><summary><i>Previously, efloden (Earl Floden) wrote…</i></summary><blockquote>

for example here: https://github.com/zanata/zanata-platform/blob/master/server/zanata-frontend/src/app/actions/review-actions.js#L40
</blockquote></details>

ah. At this point, I'm only replacing the one that is causing storybook to break. I still prefer to use JSON library whenever we can. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-platform/715)*

seanf:
> ah. At this point, I'm only replacing the one that is causing storybook to break. I still prefer to use JSON library whenever we can.

:+1: I think the replacements tend to be slower than the built-in JS functions, as well as being less standard. eg https://jsperf.com/lodash-keys-vs-object-keys (`Object.keys` is twice as fast on my machine in Chrome 64.)

Not that performance is the highest priority. Sticking to standard, better tested code is probably a more important factor.

spathare21:
@aeng still sees the same error When running ./build -ciQ --server in zanata-platform

seanf:
@spathare21 it turns out that those warnings don't mean much. The failure to run storybook was the real problem.

spathare21:
@seanf ohh ok let me check it then

spathare21:
Verified the storybook is working fine. only that error/warning message still shows while build

PR #744, opened 2018-03-09
fix(ZNTA-193): Prevent user exceeding system filename length

djansen-redhat:
JIRA issue URL: https://zanata.atlassian.net/browse/ZNTA-193

Prevent a user specifying a file path exceeding the standard 255 characters.

## Checklist

- JIRA link
- Check target branch

----

### DIFF
--- a/server/zanata-war/src/main/webapp/resources/zanata/multi-file-upload.xhtml
+++ b/server/zanata-war/src/main/webapp/resources/zanata/multi-file-upload.xhtml
@@ -114,6 +114,9 @@ site: http://www.fsf.org.
           <li data-key="jsf.upload.TooManyFiles">
             #{msgs['jsf.upload.TooManyFiles']}
           </li>
+          <li data-key="jsf.upload.FilenameIsTooLarge">
+            #{msgs['jsf.upload.FilenameIsTooLarge']}
+          </li>
           <li data-key="jsf.upload.SessionTimedOut">
             #{msgs['jsf.upload.SessionTimedOut']}
           </li>
@@ -216,7 +219,7 @@ site: http://www.fsf.org.
               <div id="#{cc.clientId}-fileupload-advanced"
                 class="l--pad-all-half bg--pop-higher is-hidden">
                 <label for="#filepath">#{msgs['jsf.upload.FilePath']}</label>
-                <input type="text" id="filepath" name="filepath"/>
+                <input type="text" id="filepath" maxlength="250" name="filepath"/>
                 <!-- Source language is currently ignored by the server.
                 <label for="filelang">Source Language</label>
                 <select name="filelang" id="filelang">
--- a/server/services/src/main/java/org/zanata/service/impl/DocumentServiceImpl.java
+++ b/server/services/src/main/java/org/zanata/service/impl/DocumentServiceImpl.java
@@ -44,6 +44,7 @@ import org.zanata.dao.ProjectIterationDAO;
 import org.zanata.events.DocStatsEvent;
 import org.zanata.events.DocumentLocaleKey;
 import org.zanata.events.DocumentUploadedEvent;
+import org.zanata.exception.ZanataServiceException;
 import org.zanata.model.type.WebhookType;
 import org.zanata.i18n.Messages;
 import org.zanata.lock.Lock;
@@ -175,6 +176,10 @@ public class DocumentServiceImpl implements DocumentService {
             // TODO check that entity name matches id parameter
             document = new HDocument(sourceDoc.getName(),
                     sourceDoc.getContentType(), hLocale);
+            if(document.getPath().length() + document.getName().length() > 255) {
+                throw new ZanataServiceException(
+                        msgs.get("jsf.upload.FilenameIsTooLarge"));
+            }
             document.setProjectIteration(hProjectIteration);
             hProjectIteration.getDocuments().put(docId, document);
             document = documentDAO.makePersistent(document);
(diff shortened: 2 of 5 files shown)


djansen-redhat:
@efloden I can't seem to get the value to display correctly. It should indicate failures, rather than 'Uploaded 1 of 1'.
screenshot from 2018-03-09 10-47-04

seanf:
@djansen-redhat Why 255? or is it 250? Which filesystem are you thinking of?

djansen-redhat:
@seanf seems I may have misinterpreted what I was reading.
255 is the filename limit in linux, but _path is 4096_. It may actually be the database that is placing this hard limit on the path length.

server/zanata-war/src/main/webapp/resources/zanata/multi-file-upload.xhtml
```diff
@@ -216,7 +219,7 @@ site: http://www.fsf.org.
               <div id="#{cc.clientId}-fileupload-advanced"
                 class="l--pad-all-half bg--pop-higher is-hidden">
                 <label for="#filepath">#{msgs['jsf.upload.FilePath']}</label>
-                <input type="text" id="filepath" name="filepath"/>
+                <input type="text" id="filepath" maxlength="250" name="filepath"/>
```
seanf: @djansen-redhat Why is this one only 250?

seanf:
@djansen-redhat Yes, it probably is the DB column limits. The only time filesystem limits should affect file uploads is when uploading "raw" files, which are currently stored in the "zanata documents directory", but I don't think we use the original filename to create those files anyway.

seanf:
@djansen-redhat If you capitalise ZNTA in the branch name, JIRA should be able to link to the branch from the ZNTA issue.

server/zanata-war/src/main/webapp/resources/zanata/multi-file-upload.xhtml
djansen-redhat: 250 + filename of 4 characters (oops :laughing: )

djansen-redhat:
@seanf so I guess the db limit (confirmed varchar255) was arbitrarily chosen.

Is the fix now 
- increase that limit to something like 4098 - 255, 
- a considerable limit with a "yeah we don't support anything beyond this'
- 4097 characters with smarts to work out if files breach that limit before upload?

seanf:
@djansen-redhat If I understand correctly, the 4095 limit isn't an absolute limit on path names imposed by the filesystem, but a common limit imposed by applications or their libraries (eg the C standard library). So it's possible to create files which have longer absolute paths, but many applications will have trouble accessing such files, especially if they happen to use absolute paths. But the limit of 255 per path segment is a hard limitation common to typical filesystems. (The 4096 limit is for C style strings, so you have to subtract 1 for the `nul` at the end.)

I think zanata-cli mostly uses relative paths, but perhaps not always. Our Maven plugin probably tends towards absolute paths.

I think it would be reasonable for us to impose a limit (in code) of 4095 on HDocument.fullPath (directory plus filename, AKA docId) and a limit of 255 on the filename. We don't store fullPath directly in the DB, so perhaps we should set the maximum to 4095 for "path" and 255 for "name" (in the DB). Note: this may add up to more than 4095 in some cases, but that's what the "in code" limit will be for.

I think we can change/impose the limits in code by using the validator `@Size` on '.name' and 'docId'.

Whatever limit we choose, it's possible that a particular OS/filesystem/client may have smaller limits (thus leading to some sort of runtime error), so I don't think we should worry about trying to prevent it entirely. Better just to have **some** limits in Zanata server, and try to avoid making them the limiting factor.

seanf:
@djansen-redhat 

> @seanf so I guess the db limit (confirmed varchar255) was arbitrarily chosen.

I think so. I'm not sure if anyone (other than you) has run into the limit yet though. (Some Java programmer might, with .properties in a deep source tree.)

> Is the fix now 
> - increase that limit to something like 4098 - 255, 
> - a considerable limit with a "yeah we don't support anything beyond this'
> - 4097 characters with smarts to work out if files breach that limit before upload?

If we were to increase the limits, I would try to match Linux's limits (255 and 4095).

> - (edit) keep what we have now and just let them fail but with a better error

I'm never sure what to do about redundant validation problems like this. (For example, GWT was designed to let us re-use the server-side annotation-based bean validators on the client side.) If you do nothing in the client, the limit will still be enforced somewhere, but you might not get a nice error message, and it might not happen until you commit. Probably leading to a very ugly error message, depending on how we handle it.

I don't really like the idea of redundantly specifying the limits so many times (eg how to make sure they stay in sync?):

1. In the DB schema (AND the Hibernate/JPA annotations like `@Column length`, because we can't sync them with Liquibase yet.)
2. In the Hibernate entity's bean validation annotations like `@Size`
3. On the client-side (eg JavaScript for the browser, or zanata-cli)

But I suppose it's the only way to give timely error messages to the user. It seems like we ought be passing the limits from the bean validation annotations down to the client somehow (eg shared constants, or runtime queries against the annotations). But I'm sure that's a much bigger problem than you wanted to tackle. I suppose it would suffice to put comments in each location, to help us keep them in sync if they ever change.

I think if we put the filename/path limits as constants into zanata-common-api, we should be able to use them from the annotations `@Column length`, `@Size` and also from the zanata-cli code. However, it's probably okay for zanata-cli just to ignore the limits, and handle the validation error when it comes back from the server. (In practice, the client will probably be unable to generate excessive filenames due to filesystem limitations, assuming we increase Zanata's limits to match Linux. Unless the client is using filename mapping to complicate things...) That's assuming the server is returning a meaningful error, and not a 500 code or something.

PR #763, opened 2018-03-27
fix(ZNTA-363): Hide email addresses in translations request

djansen-redhat:
JIRA issue URL: https://zanata.atlassian.net/browse/ZNTA-363

Hide email addresses in translations REST requests, if the user is not an admin and the 'display emails' config is not set.

## Checklist

- JIRA link
- Check target branch

----

### DIFF
--- a/server/services/src/main/java/org/zanata/rest/service/ResourceUtils.java
+++ b/server/services/src/main/java/org/zanata/rest/service/ResourceUtils.java
@@ -43,6 +43,7 @@ import org.zanata.common.ContentState;
 import org.zanata.common.LocaleId;
 import org.zanata.common.MergeType;
 import org.zanata.common.ResourceType;
+import org.zanata.config.DatabaseBackedConfig;
 import org.zanata.dao.LocaleDAO;
 import org.zanata.model.HDocument;
 import org.zanata.model.HLocale;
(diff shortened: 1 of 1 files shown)


djansen-redhat:
@aeng not sure if this is the right approach, perhaps a bit blunt

djansen-redhat:
@seanf I don't _believe_ so, it should only be for the json response. But a good question that I'll look into.

seanf:
Marking as "on hold" until we decide what to do about the Gettext problem.

server/services/src/main/java/org/zanata/rest/service/ResourceUtils.java
```diff
@@ -1392,13 +1398,17 @@ public boolean transferToTranslationsResource(TranslationsResource transRes,
         boolean found = this.transferToTranslationsResourceExtensions(document,
                 transRes.getExtensions(true), enabledExtensions, locale,
                 hTargets);
+        boolean showEmail = config.isDisplayUserEmail() || identity.hasRole("admin");
```
djansen-redhat: Admin should not be able to see it either

seanf: As discussed with @aeng, admins may need extra information like that to do their admin-type things. The main thing is to avoid showing the email address to the public.

PR #770, opened 2018-04-05
fix(ZNTA-568): Provide upload failure details to the user

djansen-redhat:
JIRA issue URL: https://zanata.atlassian.net/browse/ZNTA-568

Where possible, pass the exception reason to the user so it
is easier to debug a failed upload.

## Checklist

- JIRA link
- Check target branch

----

### DIFF
--- a/common/zanata-adapter-po/src/main/java/org/zanata/adapter/po/PoReader2.java
+++ b/common/zanata-adapter-po/src/main/java/org/zanata/adapter/po/PoReader2.java
@@ -96,7 +96,7 @@ public class PoReader2 {
 
             if (message.isHeader()) {
                 if (headerFound)
-                    throw new IllegalStateException("multiple headers found");
+                    throw new PoReaderException("Multiple headers found");
                 headerFound = true;
 
                 // add target header data
@@ -208,7 +208,7 @@ public class PoReader2 {
 
             if (message.isHeader()) {
                 if (headerFound)
-                    throw new IllegalStateException("multiple headers found");
+                    throw new PoReaderException("Multiple headers found");
                 headerFound = true;
 
                 // store POT data
--- a/common/zanata-adapter-po/src/main/java/org/zanata/adapter/po/PoReader2.java
+++ b/common/zanata-adapter-po/src/main/java/org/zanata/adapter/po/PoReader2.java
@@ -96,7 +96,7 @@ public class PoReader2 {
 
             if (message.isHeader()) {
                 if (headerFound)
-                    throw new IllegalStateException("found a second header!");
+                    throw new IllegalStateException("multiple headers found");
                 headerFound = true;
 
                 // add target header data
@@ -208,7 +208,7 @@ public class PoReader2 {
 
             if (message.isHeader()) {
                 if (headerFound)
-                    throw new IllegalStateException("found a second header!");
+                    throw new IllegalStateException("multiple headers found");
                 headerFound = true;
 
                 // store POT data
--- a/common/zanata-adapter-po/src/main/java/org/zanata/adapter/po/PoReaderException.java
+++ b/common/zanata-adapter-po/src/main/java/org/zanata/adapter/po/PoReaderException.java
@@ -1,0 +1,13 @@
+package org.zanata.adapter.po;
+
+public class PoReaderException extends RuntimeException {
+
+    private static final long serialVersionUID = 1L;
+
+    public PoReaderException() {
+    }
+
+    public PoReaderException(String message) {
+        super(message);
+    }
+}
--- a/server/services/src/main/java/org/zanata/adapter/OkapiFilterAdapter.java
+++ b/server/services/src/main/java/org/zanata/adapter/OkapiFilterAdapter.java
@@ -31,6 +31,7 @@ import java.util.Map;
 import net.sf.okapi.common.Event;
 import net.sf.okapi.common.EventType;
 import net.sf.okapi.common.IParameters;
+import net.sf.okapi.common.exceptions.OkapiException;
 import net.sf.okapi.common.exceptions.OkapiIOException;
 import net.sf.okapi.common.filters.IFilter;
 import net.sf.okapi.common.filterwriter.GenericContent;
@@ -197,7 +198,7 @@ public class OkapiFilterAdapter implements FileFormatAdapter {
                     }
                 }
             }
-        } catch (OkapiIOException e) {
+        } catch (OkapiException e) {
             throw new FileFormatAdapterException("Unable to parse document", e);
         } finally {
             filter.close();
@@ -332,7 +333,7 @@ public class OkapiFilterAdapter implements FileFormatAdapter {
                     }
                 }
             }
-        } catch (OkapiIOException e) {
+        } catch (OkapiException e) {
             throw new FileFormatAdapterException(
                     "Unable to parse translation file", e);
         } finally {
--- a/server/services/src/main/java/org/zanata/file/SourceDocumentUpload.java
+++ b/server/services/src/main/java/org/zanata/file/SourceDocumentUpload.java
@@ -28,6 +28,7 @@ import java.io.InputStream;
 import java.io.Serializable;
 import javax.annotation.Nonnull;
 import javax.enterprise.context.Dependent;
+import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.Response.Status;
 import com.google.common.collect.Sets;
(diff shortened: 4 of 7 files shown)


djansen-redhat:
QA notes:
- [ ] try to push some broken files. "Internal server error" should not show, but something like "... missing thing on line 5"
- [ ] I've tried a pot file with duplicate headers, duplicate IDs and a Json file with garbage in it. Tests may cover more broken ideas.
- [ ] Upload fail numbers should be accurate, and the UI should not break on messages.

djansen-redhat:
*[server/services/src/main/java/org/zanata/file/SourceDocumentUpload.java, line 291 at r2](https://beta.reviewable.io/reviews/zanata/zanata-platform/770#-L9JcTP05supwiEq6DUs:-L9NR9Ta2ca0Bjj0wVa1:b-nh50k8) ([raw file](https://github.com/zanata/zanata-platform/blob/1a18e706979b93d45896b87f8ff5e3b58d7bfcfa/server/services/src/main/java/org/zanata/file/SourceDocumentUpload.java#L291)):*

Looks like this is a the source of some of our upload problems. Why would we want to return INTERNAL_SERVER_ERROR deliberately?

Normally I would say we shouldn't catch ZanataServiceException or WebApplicationException, because I think they both encode their own HTTP status codes (which are then handled by a generic exception handler somewhere else).

In this case, we might be better off catching SecurityException and returning a suitable security-related HTTP code, say 403, and letting the other two exceptions go through to the wicket keeper.
</blockquote></details>

I agree with the first part, but perhaps there's a lack of a "wicket keeper"  as WebApplicationException slipped through as a 500 error.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/zanata/zanata-platform/770)*

djansen-redhat:
*[server/services/src/main/java/org/zanata/rest/service/ResourceUtils.java, line 167 at r2](https://beta.reviewable.io/reviews/zanata/zanata-platform/770#-L9Jd2y13cNU4hw0tSzs:-L9NRfgeD3olkn4yUzpy:b-56qqdf) ([raw file](https://github.com/zanata/zanata-platform/blob/cc5ddcdb55ca7d0be7a23afb311fd099fe5eae08/server/services/src/main/java/org/zanata/rest/service/ResourceUtils.java#L167)):*

Why remove the id? Kind of hard for the end user to fix it without that information.
</blockquote></details>

The id is a hash of the tf content, and is basically useless to the user. I considered returning the offending content, but that was a significant risk. Maybe a truncated clip of the content?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/zanata/zanata-platform/770)*

seanf:
> The id is a hash of the tf content, and is basically useless to the user. I considered returning the offending content, but that was a significant risk. Maybe a truncated clip of the content?

That depends on the file type. And even if the user can't read the `tfId`, it gives **us** a chance at diagnosis when it shows up in a bug report.

- For Properties and XLIFF that `tfId` is not a hash, and is the best way of identifying a text flow. 
- For Gettext, it is a hash. The best human-readable option would be content[0] plus htextflow.getPotEntryData().getContext() (although it might be necessary to truncate content[0] in some cases)
- For Okapi adapters, it depends on the file type again - could be a hash, might be a useful identifier

Some time we should try to create an abstraction which returns the best human-readable identifier for any given type of text flow. Plus an abbreviated version of same.

But for now, I suggest including the `tfId` (which _might_ be a hash) plus you could throw in ShortString.shorten(content[0]).

seanf:
> I agree with the first part, but perhaps there's a lack of a "wicket keeper" as WebApplicationException slipped through as a 500 error.

Do you have the stack trace? We must have some code which is throwing a generic WebApplicationException without a code (thus defaulting to 500). We should probably put a stop to that.

djansen-redhat:
*[server/services/src/main/java/org/zanata/service/impl/TranslationFileServiceImpl.java, line 278 at r2](https://beta.reviewable.io/reviews/zanata/zanata-platform/770#-L9JdgGb18CSYp6jV38B:-L9OEgzBB9RdqGVNk1zX:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/1a18e706979b93d45896b87f8ff5e3b58d7bfcfa/server/services/src/main/java/org/zanata/service/impl/TranslationFileServiceImpl.java#L278)):*

If we change PoReader2, perhaps we could avoid the need to catch IllegalStateException, which is kind of nasty. We should probably introduce a new exception in PoReader2's package, eg PoReaderException which extends RuntimeException.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/zanata/zanata-platform/770)*

djansen-redhat:
*[server/zanata-war/src/test/java/org/zanata/rest/service/raw/TranslationResourceRestITCase.java, line 342 at r2](https://beta.reviewable.io/reviews/zanata/zanata-platform/770#-L9JfUjY82xYT7UkKBc2:-L9OEkTQ2IO0hJh77DYp:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/1a18e706979b93d45896b87f8ff5e3b58d7bfcfa/server/zanata-war/src/test/java/org/zanata/rest/service/raw/TranslationResourceRestITCase.java#L342)):*

As I said, I think it's useful to have the TF id in the error message. Perhaps we could use a regex like "duplicate.*tf1" for the test here.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/zanata/zanata-platform/770)*

djansen-redhat:
*[server/services/src/main/java/org/zanata/rest/service/ResourceUtils.java, line 167 at r2](https://beta.reviewable.io/reviews/zanata/zanata-platform/770#-L9Jd2y13cNU4hw0tSzs:-L9OEyRgAtw7j-1n2ylm:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/cc5ddcdb55ca7d0be7a23afb311fd099fe5eae08/server/services/src/main/java/org/zanata/rest/service/ResourceUtils.java#L167)):*
<details><summary><i>Previously, djansen-redhat (Damian Jansen) wrote…</i></summary><blockquote>

The id is a hash of the tf content, and is basically useless to the user. I considered returning the offending content, but that was a significant risk. Maybe a truncated clip of the content?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/zanata/zanata-platform/770)*

djansen-redhat:
I'll have to checkout master and investigate.

---

Review status: 1 of 7 files reviewed at latest revision, 5 unresolved discussions, some commit checks broke.

---

*[server/services/src/main/java/org/zanata/service/impl/TranslationFileServiceImpl.java, line 189 at r2](https://beta.reviewable.io/reviews/zanata/zanata-platform/770#-L9JdQ1TBmciP0_9ObPg:-L9OF-pTDrqnKvaG2DzZ:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/1a18e706979b93d45896b87f8ff5e3b58d7bfcfa/server/services/src/main/java/org/zanata/service/impl/TranslationFileServiceImpl.java#L189)):*

I think the old error message text is more informative - it's an error during parsing.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/zanata/zanata-platform/770#-:-L9OF47rBBskWfOH-Xre:b-254uka)*

seanf:
Reviewed 6 of 6 files at r4.

---

*[server/services/src/main/java/org/zanata/file/SourceDocumentUpload.java, line 291 at r2](https://beta.reviewable.io/reviews/zanata/zanata-platform/770#-L9JcTP05supwiEq6DUs:-L9OeWdH1Kn-mESeCnrd:bbgqo9) ([raw file](https://github.com/zanata/zanata-platform/blob/1a18e706979b93d45896b87f8ff5e3b58d7bfcfa/server/services/src/main/java/org/zanata/file/SourceDocumentUpload.java#L291)):*
<details><summary><i>Previously, djansen-redhat (Damian Jansen) wrote…</i></summary><blockquote>

I agree with the first part, but perhaps there's a lack of a "wicket keeper"  as WebApplicationException slipped through as a 500 error.
</blockquote></details>

Both ZanataServiceException and WebApplicationException can carry HTTP status codes, and I don't like the idea of throwing away that information here, but unfortunately both those exceptions default to the 500 code, which means they are often thrown with 500, eg when there is a parsing error. The fact that we do that for badly formed user input is a bug.

Maybe we need to go through the places which throw these exceptions and ensure that they set suitable codes. Then again, perhaps putting the code there was a mistake all along, since it makes an assumption about how the code will be called (even if that assumption seems to be true).

So perhaps we need to go through all such places which are parsing files and have them throw some sort of ParsingException [instead of ZanataServiceException/WebApplicationException], so that they can all be converted to BAD_REQUEST here (or better still in the ExceptionMapper/ExceptionHandler wicket keeper - I think it's safe to say that any ParsingException while processing an HTTP request represents bad input and thus BAD_REQUEST is appropriate).

There aren't that many places in the code which match the regex `(throw|new) (ZanataServiceException|WebApplicationException)` especially if you concentrate on the files which are related to parsing.

Let me think about it for a bit.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/zanata/zanata-platform/770)*

MWFIAE:
Any news on this? Getting an error during upload but no clue what exactly is wrong 🙏

PR #774, opened 2018-04-05
ZNTA-2472: enable includeFuzzy pull for .properties file type

JoeChester:
JIRA issue URL: [ZNTA-2472](https://zanata.atlassian.net/browse/ZNTA-2472)

respect the "ignoreFuzzy" option for .properties-Pulls (cli and Maven plugin)

### DIFF
--- a/client/zanata-client-commands/src/main/java/org/zanata/client/commands/pull/PropertiesStrategy.java
+++ b/client/zanata-client-commands/src/main/java/org/zanata/client/commands/pull/PropertiesStrategy.java
@@ -63,10 +63,11 @@ public class PropertiesStrategy extends AbstractPullStrategy {
             LocaleMapping localeMapping, TranslationsResource targetDoc)
             throws IOException {
         boolean createSkeletons = getOpts().getCreateSkeletons();
+        boolean includeFuzzy = getOpts().getIncludeFuzzy();
         File transFileToWrite = getTransFileToWrite(docName, localeMapping);
         PropWriter.writeTranslationsFile(createSkeletons ? doc : null,
                 targetDoc, transFileToWrite, PropWriter.CHARSET.Latin1,
-                createSkeletons);
+                createSkeletons, includeFuzzy);
 
         return null;
     }
--- a/client/zanata-client-commands/src/main/java/org/zanata/client/commands/pull/UTF8PropertiesStrategy.java
+++ b/client/zanata-client-commands/src/main/java/org/zanata/client/commands/pull/UTF8PropertiesStrategy.java
@@ -51,10 +51,11 @@ public class UTF8PropertiesStrategy extends PropertiesStrategy {
             LocaleMapping localeMapping, TranslationsResource targetDoc)
             throws IOException {
         boolean createSkeletons = getOpts().getCreateSkeletons();
+        boolean includeFuzzy = getOpts().getIncludeFuzzy();
         File transFileToWrite = getTransFileToWrite(docName, localeMapping);
         PropWriter.writeTranslationsFile(doc, targetDoc,
             transFileToWrite,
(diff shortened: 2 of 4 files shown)


aeng:
We are currently experiencing a build issue (https://github.com/victims/maven-security-versions/issues/4).

Please do a cherry-pick on the following commit to bypass the issue:
https://github.com/zanata/zanata-platform/pull/775/commits/0f1c533ec02ff972a3bbb2345679767c8908637b

seanf:
Reviewed 4 of 4 files at r1.

---

*[common/zanata-adapter-properties/src/main/java/org/zanata/adapter/properties/PropWriter.java, line 162 at r1](https://beta.reviewable.io/reviews/zanata/zanata-platform/774#-L9OBlp5802xb2O3z7nx:-L9OBlp5802xb2O3z7ny:b3azfxl) ([raw file](https://github.com/zanata/zanata-platform/blob/8e2de911574f4700970ee22590bd809c8d0ad92f/common/zanata-adapter-properties/src/main/java/org/zanata/adapter/properties/PropWriter.java#L162)):*
> ```Java
>         if (target == null || (!target.getState().isTranslated() && !includeFuzzy)
>                 || target.getContents() == null
>                 || target.getContents().size() == 0) {
> ```

This logic needs to be simplified. I can't tell if it's right or wrong, and I can't understand how `includeFuzzy` is interacting with `createSkeletons` and the translation state.

According to IntelliJ, this is essentially equivalent, and slightly more readable IMHO:

        if (target == null
                || !(includeFuzzy || target.getState().isTranslated())
                || target.getContents() == null
                || target.getContents().size() == 0) {
            // don't save fuzzy or empty values unless includeFuzzy is set
            if (createSkeletons) {
                targetProp.setProperty(resId, "");
            }
            return;
        }

(I made a slight semantic change by swapping the short-circuit evaluation around.)

I'd like to split up the `if` statement, but so far I can't find a way to do that without duplicating the body.

It's important that we have a unit test for this before we mess with this logic. Could you please add a test which covers the combinations of `createSkeletons` and `includeFuzzy`, as well as the states `Translated`, `NeedReview` and `New` ?

---

*[server/services/src/main/java/org/zanata/adapter/GenericPropertiesAdapter.java, line 125 at r1](https://beta.reviewable.io/reviews/zanata/zanata-platform/774#-L9OASHn9CgB7PQG7G36:-L9OASHn9CgB7PQG7G37:br1p4or) ([raw file](https://github.com/zanata/zanata-platform/blob/8e2de911574f4700970ee22590bd809c8d0ad92f/server/services/src/main/java/org/zanata/adapter/GenericPropertiesAdapter.java#L125)):*
> ```Java
>         // write source string with empty translation
>         boolean createSkeletons = true;
>         boolean includeFuzzy = true;
> ```

We should never include fuzzy strings in Properties files by default.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/zanata/zanata-platform/774)*

seanf:
@JoeChester You shouldn't need to do that cherry pick now. The fix has been merged to master.

rolandort:
Any progress on this issue? It would be great to have `--include-fuzzy` switch working as soon as possible for any project type. 🙏 (we use utf8properties)

tholu:
+1 for making some progress here. Why are existing pull requests not reviewed and merged? Is Zanata not actively developed anymore?

Perneel:
> +1 for making some progress here. Why are existing pull requests not reviewed and merged? Is Zanata not actively developed anymore?

I wonder as well...

carlosmunoz:
the team previously working on Zanata has moved on to other things. There are some Zanata forks which might be seeing more activity. For example: https://github.com/porscheinformatik/zanata-platform

PR #775, opened 2018-04-05
Update XML libs

seanf:


seanf:
Included in https://github.com/zanata/zanata-platform/pull/973

PR #794, opened 2018-04-20
fix(ZNTA-2387): Add email and name to user search query

djansen-redhat:
JIRA issue URL: https://zanata.atlassian.net/browse/ZNTA-2387

Admin can search for users via display name, email or username.

## Checklist

- JIRA link
- Check target branch

----

### DIFF
--- a/server/services/src/main/java/org/zanata/dao/AccountDAO.java
+++ b/server/services/src/main/java/org/zanata/dao/AccountDAO.java
@@ -17,7 +17,6 @@ import java.security.SecureRandom;
 import java.util.List;
 import java.util.Optional;
 
-import org.apache.commons.lang3.ObjectUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.hibernate.Criteria;
 import org.hibernate.Query;
@@ -137,10 +136,10 @@ public class AccountDAO extends AbstractDAOImpl<HAccount, Long> {
     }
 
     public int getUserCount(String filter) {
-        Query query = createFilteredQuery(
+        return ((Long) createFilteredQuery(
                 "select count(*) from HAccount acc ", filter)
-                .setComment("accountDAO.getUserCount");
-        return ObjectUtils.firstNonNull((Long) query.uniqueResult(), 0).intValue();
+                .setComment("accountDAO.getUserCount").uniqueResult())
+                .intValue();
     }
 
     private Query createFilteredQuery(String queryBase, String filter) {
--- a/server/services/src/main/java/org/zanata/dao/AccountDAO.java
+++ b/server/services/src/main/java/org/zanata/dao/AccountDAO.java
@@ -126,35 +126,34 @@ public class AccountDAO extends AbstractDAOImpl<HAccount, Long> {
     }
 
     public List<String> getUserNames(String filter, int offset, int maxResults) {
-        Query query = createQuery(new StringBuilder(
-                "select distinct acc.username from HAccount acc "), filter);
-        query.setMaxResults(maxResults);
-        query.setFirstResult(offset);
-        query.setComment("accountDAO.getUserNames");
+        Query query = createFilteredQuery(
+                "select distinct acc.username from HAccount acc ", filter)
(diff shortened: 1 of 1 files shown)


djansen-redhat:
@seanf I've dropped the Kotlin convert as it opened a bit of a rabbit hole, but saved it elsewhere. I'll try it after this goes through.

seanf:
Reviewed 1 of 1 files at r2.

---

*[server/services/src/main/java/org/zanata/dao/AccountDAO.java, line 130 at r2](https://reviewable.io/reviews/zanata/zanata-platform/794#-LBIsGd5-G-U-sUNYFcz:-LBIsGd5-G-U-sUNYFd-:be5uu45) ([raw file](https://github.com/zanata/zanata-platform/blob/55a3f19cdf70c0ccdd3ac20e5b07def009170783/server/services/src/main/java/org/zanata/dao/AccountDAO.java#L130)):*
> ```Java
>     public List<String> getUserNames(String filter, int offset, int maxResults) {
>         Query query = createFilteredQuery(
>                 "select distinct acc.username from HAccount acc ", filter)
> ```

"distinct" seems like overkill if username has a unique index, but okay.

---

*[server/services/src/main/java/org/zanata/dao/AccountDAO.java, line 143 at r2](https://reviewable.io/reviews/zanata/zanata-platform/794#-LBIsK7q5CcpnSdD4BqA:-LBIsK7q5CcpnSdD4BqB:bn61bne) ([raw file](https://github.com/zanata/zanata-platform/blob/55a3f19cdf70c0ccdd3ac20e5b07def009170783/server/services/src/main/java/org/zanata/dao/AccountDAO.java#L143)):*
> ```Java
>                 "select count(*) from HAccount acc ", filter)
>                 .setComment("accountDAO.getUserCount");
>         return ObjectUtils.firstNonNull((Long) query.uniqueResult(), 0).intValue();
> ```

I'm pretty sure "select count(*)" is guaranteed to return a value. But if it doesn't, I think a noisy exception beats silent misbehaviour. So firstNonNull shouldn't be needed.

Ah, I see that the old code was doing something similar. Best to change it though, I think.

---


*Comments from [Reviewable](https://reviewable.io/reviews/zanata/zanata-platform/794)*

djansen-redhat:
*[server/services/src/main/java/org/zanata/dao/AccountDAO.java, line 143 at r2](https://reviewable.io/reviews/zanata/zanata-platform/794#-LBIsK7q5CcpnSdD4BqA:-LBJW7W-3gqZjNs1-NCr:bb6un1j) ([raw file](https://github.com/zanata/zanata-platform/blob/55a3f19cdf70c0ccdd3ac20e5b07def009170783/server/services/src/main/java/org/zanata/dao/AccountDAO.java#L143)):*

I'm pretty sure "select count(*)" is guaranteed to return a value. But if it doesn't, I think a noisy exception beats silent misbehaviour. So firstNonNull shouldn't be needed.

Ah, I see that the old code was doing something similar. Best to change it though, I think.
</blockquote></details>

Done

---


*Comments from [Reviewable](https://reviewable.io/reviews/zanata/zanata-platform/794)*

djansen-redhat:
*[server/services/src/main/java/org/zanata/dao/AccountDAO.java, line 130 at r2](https://reviewable.io/reviews/zanata/zanata-platform/794#-LBIsGd5-G-U-sUNYFcz:-LBJWHm99rhvRyYJo2Ay:b-x4wifx) ([raw file](https://github.com/zanata/zanata-platform/blob/55a3f19cdf70c0ccdd3ac20e5b07def009170783/server/services/src/main/java/org/zanata/dao/AccountDAO.java#L130)):*

"distinct" seems like overkill if username has a unique index, but okay.
</blockquote></details>

I wasn't sure if I'd get a duplicated result with the OR

---


*Comments from [Reviewable](https://reviewable.io/reviews/zanata/zanata-platform/794)*

spathare21:
@djansen-redhat verified search works fine but looks like build failed for codecov

djansen-redhat:
Codecov is loopy.

PR #811, opened 2018-05-09
ZNTA-2492 erase a user and its data

huangp:
JIRA issue URL: https://zanata.atlassian.net/browse/ZNTA-2492

EU GDPR (General Data Protection Regulation) compliance, Data Protection Act 1998, et al

## Checklist

- JIRA link
- Check target branch

----


seanf:
Reviewed 21 of 21 files at r1.

---

*[server/services/src/main/java/org/zanata/action/UserAction.java, line 121 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7Qp5rDy4gxEeJ-3AQ:-LC7Qp5rDy4gxEeJ-3AR:b-hovd82) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/services/src/main/java/org/zanata/action/UserAction.java#L121)):*
> ```Java
>                     .filter(t -> t instanceof ConstraintViolationException)
>                     .findAny();
>             if (constraintEx.isPresent()) {
> ```

I think you could use causalChain.anyMatch(lambda) here to save a few lines.

---

*[server/services/src/main/java/org/zanata/exception/handler/DeleteUserExceptionHandler.java, line 38 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7N4TI2c4mdLUq3UUk:-LC7N4TI2c4mdLUq3UUl:b-cswjeu) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/services/src/main/java/org/zanata/exception/handler/DeleteUserExceptionHandler.java#L38)):*
> ```Java
>     public void handleException(@Handles ExceptionEvent<DeleteUserException> event) {
>         handle(event, LogLevel.Warn, FacesMessage.SEVERITY_WARN,
>                 "jsf.UserManager.delete.constraintViolation.error");
> ```

The specific nature of this error message suggests that a better name for the exceptino would be DeleteUserConstraintException.

---

*[server/services/src/main/java/org/zanata/file/DocumentStorage.java, line 35 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7NHfo32IK9vQWATFx:-LC7NHfo32IK9vQWATFy:bnvr6v0) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/services/src/main/java/org/zanata/file/DocumentStorage.java#L35)):*
> ```Java
> @Retention(RUNTIME)
> @Target({METHOD, FIELD, PARAMETER})
> public @interface DocumentStorage {
> ```

This name isn't understandable as is. I think it would better named as DocumentStorageDir(ectory) or else with an explanatory javadoc. 

Also we don't want to confuse it with the top-level document storage location (since it seems to be the "documents" sub-directory), so some sort of javadoc is probably necessary either way.

---

*[server/services/src/main/java/org/zanata/service/UserAccountService.java, line 86 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7Nv3I-wkXIcMY9qQw:-LC7Nv3I-wkXIcMY9qQx:b-i9ffis) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/services/src/main/java/org/zanata/service/UserAccountService.java#L86)):*
> ```Java
> 
>     /**
>      * Anonymize the user and its data.
> ```

This might be better: 

    Anonymise the user by erasing associated _personal_ data.

---

*[server/services/src/main/java/org/zanata/service/impl/UserAccountServiceImpl.java, line 199 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7VSb77HBXaHmwaeOJ:-LC7VSb77HBXaHmwaeOK:b-2kirmh) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/services/src/main/java/org/zanata/service/impl/UserAccountServiceImpl.java#L199)):*
> ```Java
>         if (account == null) {
>             log.warn("can not find account by username {}", username);
>             return;
> ```

Shouldn't we return this error to the user (via an exception)? They won't see the log.

---

*[server/services/src/main/java/org/zanata/service/impl/UserAccountServiceImpl.java, line 213 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7XwO2ECqtIfIW6Wv3:-LC7XwO2ECqtIfIW6Wv4:b-99p6b7) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/services/src/main/java/org/zanata/service/impl/UserAccountServiceImpl.java#L213)):*
> ```Java
>         person.getLanguageTeamMemberships()
>                 .forEach(membership -> session.delete(membership));
>         // remove from version group
> ```

For a community instance like translate.zanata.org, we should indeed purge single-maintainer VersionGroups, Projects, Documents, RawDocuments.

But we shouldn't delete these artifacts for corporate Zanata instances, where projects and documents need to outlive users, (even if that means Admin will have to appoint a new maintainer at some point).

We probably need a global setting so that admin can set the document erasure policy (or a system property to be configured by Ops).

---

*[server/services/src/main/java/org/zanata/service/impl/UserAccountServiceImpl.java, line 230 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7WPW61BXPZIyYll7R:-LC7WPW61BXPZIyYll7S:b-cudbzd) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/services/src/main/java/org/zanata/service/impl/UserAccountServiceImpl.java#L230)):*
> ```Java
> 
>         });
>         session.flush();
> ```

What's all the manual flushing for?

---

*[server/services/src/main/java/org/zanata/service/impl/UserAccountServiceImpl.java, line 232 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7WywEDtlCrEn1YE6Q:-LC7Wu0KEJg_fhONXw0d:b-vts57q) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/services/src/main/java/org/zanata/service/impl/UserAccountServiceImpl.java#L232)):*
> ```Java
>         session.flush();
>         // remove from project membership
>         // delete projects that is solely maintained by this user
> ```

What if the project has other relationships, eg language team coordinators?

---

*[server/services/src/main/java/org/zanata/service/impl/UserAccountServiceImpl.java, line 276 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7WDf6CArWoe2mlTiO:-LC7WDf6CArWoe2mlTiP:btte9z9) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/services/src/main/java/org/zanata/service/impl/UserAccountServiceImpl.java#L276)):*
> ```Java
>             session.delete(doc);
>         });
>         session.flush();
> ```

By the time we purge all of the above relationships, it's possible that the Account could be deleted after all (although this would mean deleting the erasure metadata).

---

*[server/services/src/main/resources/messages.properties, line 1055 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7OCVjCtMf-d13RX7R:-LC7OCVjCtMf-d13RX7S:bswd7y8) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/services/src/main/resources/messages.properties#L1055)):*
> ```INI
> jsf.Enabled=Enabled
> jsf.user.delete.title=Delete user
> jsf.user.erase.title=Erase user and its data
> ```

Erase user's personal data

---

*[server/services/src/main/resources/messages.properties, line 1057 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7OFph34ZyWGz3ALQp:-LC7OFph34ZyWGz3ALQq:b-r3878k) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/services/src/main/resources/messages.properties#L1057)):*
> ```INI
> jsf.user.erase.title=Erase user and its data
> jsf.AreYouSureYouWishToDeleteThisUserThisActionCannotBeUndone=Are you sure you wish to delete this user? This action cannot be undone.
> jsf.AreYouSureYouWishToEraseThisUserThisActionCannotBeUndone=Are you sure you wish to erase this user and its data? This action cannot be undone.
> ```

"this user's personal data"

---

*[server/services/src/main/resources/messages.properties, line 1058 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7OLr08m3MF_KZZGAK:-LC7OLr08m3MF_KZZGAL:b-u5olv6) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/services/src/main/resources/messages.properties#L1058)):*
> ```INI
> jsf.AreYouSureYouWishToDeleteThisUserThisActionCannotBeUndone=Are you sure you wish to delete this user? This action cannot be undone.
> jsf.AreYouSureYouWishToEraseThisUserThisActionCannotBeUndone=Are you sure you wish to erase this user and its data? This action cannot be undone.
> jsf.UserManager.delete.constraintViolation.error=This user cannot be removed from the system. You can erase and anonymize the user and its data.
> ```

"You can erase the user's personal data instead."

---

*[server/services/src/main/resources/db/changelogs/db.changelog-4.5.xml, line 58 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7OgbAAQlj4Nqyohzy:-LC7OgbAAQlj4Nqyohzz:b-jsddmc) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/services/src/main/resources/db/changelogs/db.changelog-4.5.xml#L58)):*
> ```XML
>       <column name="erased" type="boolean" defaultValueBoolean="false" />
>       <column name="erasureDate" type="datetime" />
>       <column name="erasedBy" type="bigint" />
> ```

`erasedByAccountId` might be better

---

*[server/services/src/main/resources/db/changelogs/db.changelog-4.5.xml, line 68 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7OwIF1GMZmjKQYpAQ:-LC7Om-LDwhjHHJhv7Tx:bjz08zf) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/services/src/main/resources/db/changelogs/db.changelog-4.5.xml#L68)):*
> ```XML
>     <comment>add columns to indicate a deleted account</comment>
>     <addColumn tableName="HAccount">
>       <column name="erased" type="boolean" defaultValueBoolean="false" />
> ```

Do we really need to track `erased`, `erasedBy` and `erasureDate` in both HAccount and HPerson? That seems kind of denormalised. Every account has a person, so it seems like HPerson.erased should be enough.

I know we do erase several columns in this table, but are we going to add erased/erasedBy/erasureDate columns to the other tables (eg HCredentials, HAccountActivationKey, HAccountResetPasswordKey, HAccountOption, AllowedApp...)?

On the other hand, if we do want to track metadata about erasure in multiple classes (I see you created the Erasable interface), it would be nice if we could associate all the objects which were erased by a single erasure request.

---

*[server/services/src/main/resources/db/changelogs/db.changelog-4.5.xml, line 70 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7Ojcx8-EYbGFlX1BH:-LC7Ojcx8-EYbGFlX1BI:b-jsddmc) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/services/src/main/resources/db/changelogs/db.changelog-4.5.xml#L70)):*
> ```XML
>       <column name="erased" type="boolean" defaultValueBoolean="false" />
>       <column name="erasureDate" type="datetime" />
>       <column name="erasedBy" type="bigint" />
> ```

`erasedByAccountId` might be better

---

*[server/services/src/test/java/org/zanata/service/impl/UserAccountServiceImplTest.java, line 180 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7S4NO4ZAlvnjP_pUg:-LC7S4NO4ZAlvnjP_pUh:b-4zw5z1) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/services/src/test/java/org/zanata/service/impl/UserAccountServiceImplTest.java#L180)):*
> ```Java
>         assertThat(person.isErased()).isTrue();
>         assertThat(person.getName()).isEqualTo("Erased User");
>         assertThat(person.getLanguageMemberships()).isEmpty();
> ```

What about person.getErasedBy?

We should also test the other key/credentials tables here, assuming we decide to erase them too.

---

*[server/zanata-model/src/main/java/org/zanata/model/Eraseable.java, line 42 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7TiPQ3Dw15YF-MxIt:-LC7TiPQ3Dw15YF-MxIu:b505rmv) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/zanata-model/src/main/java/org/zanata/model/Eraseable.java#L42)):*
> ```Java
>     HAccount getErasedBy();
> 
>     void setErasedBy(HAccount erasedBy);
> ```

It would be nice if we could avoid making all these mutator methods visible everywhere these entities are visible.

---

*[server/zanata-model/src/main/java/org/zanata/model/Eraseable.java, line 44 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7SlUy7PtIal3KCafz:-LC7SlUy7PtIal3KCag-:b-d0z4ca) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/zanata-model/src/main/java/org/zanata/model/Eraseable.java#L44)):*
> ```Java
>     void setErasedBy(HAccount erasedBy);
> 
>     void eraseSelf(HAccount erasedBy);
> ```

I'm not entirely sure what this means without a javadoc, but if it erases the Erasable object, the name `erase` should be enough. 

---

*[server/zanata-model/src/main/java/org/zanata/model/HAccount.java, line 184 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7TsQn81Ol7oNsWi7c:-LC7TsQn81Ol7oNsWi7d:bh3dkv1) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/zanata-model/src/main/java/org/zanata/model/HAccount.java#L184)):*
> ```Java
>     @Temporal(TemporalType.TIMESTAMP)
>     public Date getErasureDate() {
>         return erasureDate == null ? null : new Date(erasureDate.getTime());
> ```

    // defensive copy

---

*[server/zanata-model/src/main/java/org/zanata/model/HAccount.java, line 207 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7U0r04Qbt27khSuAh:-LC7U0r04Qbt27khSuAi:bh3dkv1) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/zanata-model/src/main/java/org/zanata/model/HAccount.java#L207)):*
> ```Java
>     public void setErasureDate(Date erasureDate) {
>         this.erasureDate =
>                 erasureDate == null ? null : new Date(erasureDate.getTime());
> ```

    // defensive copy

---

*[server/zanata-model/src/main/java/org/zanata/model/HAccount.java, line 212 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7V1jiBg_MnQIaQALY:-LC7V1jiBg_MnQIaQALZ:b-fqi6c0) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/zanata-model/src/main/java/org/zanata/model/HAccount.java#L212)):*
> ```Java
>     @Override
>     public void eraseSelf(HAccount erasedBy) {
>         this.username = "_deleted_" + getId();
> ```

I think we should try to make the erased usernames and erased email addresses similar. Maybe `_erased_{counter}`?

We shouldn't expose the primary key here. We need some sort of counter (or database sequence) for erasures.

---

*[server/zanata-model/src/main/java/org/zanata/model/HAccount.java, line 215 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7PXin5xZTyQUr7yyf:-LC7PXioEcf0id6WZb22:bgkp886) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/zanata-model/src/main/java/org/zanata/model/HAccount.java#L215)):*
> ```Java
>         this.erasureDate = new Date();
>         this.erased = true;
>         this.passwordHash = null;
> ```

Perhaps we should erase the keys too: apiKey, accountActivationKey, accountResetPasswordKey

Maybe credentials, editorOptions, allowedApps too

---

*[server/zanata-model/src/main/java/org/zanata/model/HPerson.java, line 148 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7US730VMv6nnrnyaw:-LC7US730VMv6nnrnyax:bh3dkv1) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/zanata-model/src/main/java/org/zanata/model/HPerson.java#L148)):*
> ```Java
>     @Temporal(TemporalType.TIMESTAMP)
>     public Date getErasureDate() {
>         return erasureDate == null ? null : new Date(erasureDate.getTime());
> ```

    // defensive copy

---

*[server/zanata-model/src/main/java/org/zanata/model/HPerson.java, line 171 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7UUp7CC2n6T02fYs2:-LC7UUp7CC2n6T02fYs3:bh3dkv1) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/zanata-model/src/main/java/org/zanata/model/HPerson.java#L171)):*
> ```Java
>     public void setErasureDate(Date erasureDate) {
>         this.erasureDate =
>                 erasureDate == null ? null : new Date(erasureDate.getTime());
> ```

    // defensive copy

---

*[server/zanata-model/src/main/java/org/zanata/model/HPerson.java, line 177 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7U_J2COVjNKpFZK2w:-LC7U_J2COVjNKpFZK2x:b-o6k0dq) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/zanata-model/src/main/java/org/zanata/model/HPerson.java#L177)):*
> ```Java
>     public void eraseSelf(HAccount erasedBy) {
>         this.name = "Erased User";
>         this.email = "non-exist-" + getId() + "@example.com";
> ```

I think we should try to make the erased usernames and erased email addresses similar. Maybe `_erased_{counter}`?

---

*[server/zanata-model/src/main/java/org/zanata/model/HPerson.java, line 177 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7VF8IEzpp-RFUYYK2:-LC7VF8IEzpp-RFUYYK3:b-t7i55v) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/zanata-model/src/main/java/org/zanata/model/HPerson.java#L177)):*
> ```Java
>     public void eraseSelf(HAccount erasedBy) {
>         this.name = "Erased User";
>         this.email = "non-exist-" + getId() + "@example.com";
> ```

We shouldn't expose the primary key here. We need some sort of counter (or database sequence) for erasures.

---

*[server/zanata-war/src/main/webapp-jboss/WEB-INF/classes/META-INF/persistence.xml, line 90 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7PouK8wo_Q4lUamcK:-LC7PouK8wo_Q4lUamcL:b-xgwjqr) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/zanata-war/src/main/webapp-jboss/WEB-INF/classes/META-INF/persistence.xml#L90)):*
> ```XML
> 
>       
>       <property name="hibernate.show_sql" value="${env.hibernate_show_sql}" />
> ```

Is this deliberate? Was this broken before?

---

*[server/zanata-war/src/main/webapp/admin/usermanager.xhtml, line 140 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7Q73P-Mha6qaE5Bmq:-LC7Q73P-Mha6qaE5Bmr:b4fo8s0) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/zanata-war/src/main/webapp/admin/usermanager.xhtml#L140)):*
> ```HTML
>                         </ui:fragment>
>                         <ui:fragment
>                           rendered="#{identity.hasPermission('seam.user', 'delete')}">
> ```

Perhaps we should hide this if the user has been erased.

---


*Comments from [Reviewable](https://reviewable.io/reviews/zanata/zanata-platform/811)*

huangp:
Review status: all files reviewed at latest revision, 28 unresolved discussions, some commit checks failed.

---

*[server/services/src/main/java/org/zanata/action/UserAction.java, line 121 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7Qp5rDy4gxEeJ-3AQ:-LC7Z10N9puf7hOPaDqi:bs7wup3) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/services/src/main/java/org/zanata/action/UserAction.java#L121)):*

I think you could use causalChain.anyMatch(lambda) here to save a few lines.
</blockquote></details>

Done.n

---

*[server/services/src/main/java/org/zanata/exception/handler/DeleteUserExceptionHandler.java, line 38 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7N4TI2c4mdLUq3UUk:-LC7ZGRP2bNr541fXVkD:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/services/src/main/java/org/zanata/exception/handler/DeleteUserExceptionHandler.java#L38)):*

The specific nature of this error message suggests that a better name for the exceptino would be DeleteUserConstraintException.
</blockquote></details>

Done.

---

*[server/services/src/main/java/org/zanata/file/DocumentStorage.java, line 35 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7NHfo32IK9vQWATFx:-LC7_5dQ4mR8k18RSYLD:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/services/src/main/java/org/zanata/file/DocumentStorage.java#L35)):*

This name isn't understandable as is. I think it would better named as DocumentStorageDir(ectory) or else with an explanatory javadoc. 

Also we don't want to confuse it with the top-level document storage location (since it seems to be the "documents" sub-directory), so some sort of javadoc is probably necessary either way.
</blockquote></details>

Done.

---

*[server/services/src/main/java/org/zanata/service/UserAccountService.java, line 86 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7Nv3I-wkXIcMY9qQw:-LC7_PXw4wDZSnHRVKUM:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/services/src/main/java/org/zanata/service/UserAccountService.java#L86)):*

This might be better: 

    Anonymise the user by erasing associated _personal_ data.
</blockquote></details>

Done.

---

*[server/services/src/main/java/org/zanata/service/impl/UserAccountServiceImpl.java, line 199 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7VSb77HBXaHmwaeOJ:-LC7aVdL2zMh_i5B1Cdy:bmh7hz9) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/services/src/main/java/org/zanata/service/impl/UserAccountServiceImpl.java#L199)):*

Shouldn't we return this error to the user (via an exception)? They won't see the log.
</blockquote></details>

Now checking the null in UserAction.

---

*[server/services/src/main/java/org/zanata/service/impl/UserAccountServiceImpl.java, line 213 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7XwO2ECqtIfIW6Wv3:-LC7_mLR6EiIDORepB4s:b-359li9) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/services/src/main/java/org/zanata/service/impl/UserAccountServiceImpl.java#L213)):*

For a community instance like translate.zanata.org, we should indeed purge single-maintainer VersionGroups, Projects, Documents, RawDocuments.

But we shouldn't delete these artifacts for corporate Zanata instances, where projects and documents need to outlive users, (even if that means Admin will have to appoint a new maintainer at some point).

We probably need a global setting so that admin can set the document erasure policy (or a system property to be configured by Ops).
</blockquote></details>

According to what Alex has put in the jira, we need to delete them. Project and versions are soft deleted (so documents are hidden from anywhere except database). Version group can only be created by admin. I think they can be easily recreated.

---

*[server/services/src/main/java/org/zanata/service/impl/UserAccountServiceImpl.java, line 230 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7WPW61BXPZIyYll7R:-LC7c5Pr3oB3ub-zU_MH:b-77o628) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/services/src/main/java/org/zanata/service/impl/UserAccountServiceImpl.java#L230)):*

What's all the manual flushing for?
</blockquote></details>

session.delete didn't work for me. When I google it some people said you need the flush afterwards but it still won't work. In the end I did it in different ways but left the flush. If anything is not right, flush will trigger the error immediately. Otherwise it will only be flushed at the end of the transaction.

---

*[server/services/src/main/java/org/zanata/service/impl/UserAccountServiceImpl.java, line 232 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7WywEDtlCrEn1YE6Q:-LC7cpXu5LbhwhiEgfxw:b-aixvjb) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/services/src/main/java/org/zanata/service/impl/UserAccountServiceImpl.java#L232)):*

What if the project has other relationships, eg language team coordinators?
</blockquote></details>

According to JIRA we only care about maintainer. I guess language team coordinator is just like translators. They don't *own* the project.

---

*[server/services/src/main/java/org/zanata/service/impl/UserAccountServiceImpl.java, line 276 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7WDf6CArWoe2mlTiO:-LC7dE-q9QLjpvkurFcj:bbfokps) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/services/src/main/java/org/zanata/service/impl/UserAccountServiceImpl.java#L276)):*

By the time we purge all of the above relationships, it's possible that the Account could be deleted after all (although this would mean deleting the erasure metadata).
</blockquote></details>

No. We don't delete Document, TextFlow, TextFlowTarget. They are still in database but hidden from UI, REST etc.

---

*[server/services/src/main/resources/messages.properties, line 1055 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7OCVjCtMf-d13RX7R:-LC7dh3zD5NEoseSn7le:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/services/src/main/resources/messages.properties#L1055)):*

Erase user's personal data
</blockquote></details>

Done.

---

*[server/services/src/main/resources/messages.properties, line 1057 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7OFph34ZyWGz3ALQp:-LC7dp8t65baCY1B3V1R:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/services/src/main/resources/messages.properties#L1057)):*

"this user's personal data"
</blockquote></details>

Done.

---

*[server/services/src/main/resources/messages.properties, line 1058 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7OLr08m3MF_KZZGAK:-LC7dy1p-2z9UZXSiuoS:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/services/src/main/resources/messages.properties#L1058)):*

"You can erase the user's personal data instead."
</blockquote></details>

Done.

---

*[server/services/src/main/resources/db/changelogs/db.changelog-4.5.xml, line 58 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7OgbAAQlj4Nqyohzy:-LC7e3F13wLzwu4zSRuI:b-ywpqcj) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/services/src/main/resources/db/changelogs/db.changelog-4.5.xml#L58)):*

`erasedByAccountId` might be better
</blockquote></details>

we don't use id in other places. e.g. modifiedBy. And it matches the field name.

---

*[server/services/src/main/resources/db/changelogs/db.changelog-4.5.xml, line 68 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7OwIF1GMZmjKQYpAQ:-LC7edYpDyo845cYdAj4:b-qybdri) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/services/src/main/resources/db/changelogs/db.changelog-4.5.xml#L68)):*

Do we really need to track `erased`, `erasedBy` and `erasureDate` in both HAccount and HPerson? That seems kind of denormalised. Every account has a person, so it seems like HPerson.erased should be enough.

I know we do erase several columns in this table, but are we going to add erased/erasedBy/erasureDate columns to the other tables (eg HCredentials, HAccountActivationKey, HAccountResetPasswordKey, HAccountOption, AllowedApp...)?

On the other hand, if we do want to track metadata about erasure in multiple classes (I see you created the Erasable interface), it would be nice if we could associate all the objects which were erased by a single erasure request.
</blockquote></details>

Technically HAccount can exist without a HPerson (when user sign up but before activation). So definitely need flags in HAccount. Not sure about whether we need to have those in HPerson. It's in the JIRA so I did it. I guess it makes some query easier since a lot of the references are going to HPerson. e.g. select * from HTextFlow where lastModifiedBy.erased = false

---

*[server/services/src/main/resources/db/changelogs/db.changelog-4.5.xml, line 70 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7Ojcx8-EYbGFlX1BH:-LC7fU0N3sJG9WsRL8aK:b-9k2o4q) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/services/src/main/resources/db/changelogs/db.changelog-4.5.xml#L70)):*

`erasedByAccountId` might be better
</blockquote></details>

same reason as above

---

*[server/services/src/test/java/org/zanata/service/impl/UserAccountServiceImplTest.java, line 180 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7S4NO4ZAlvnjP_pUg:-LC7fj_f56KCvGSmxuMR:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/services/src/test/java/org/zanata/service/impl/UserAccountServiceImplTest.java#L180)):*

What about person.getErasedBy?

We should also test the other key/credentials tables here, assuming we decide to erase them too.
</blockquote></details>

Done.

---

*[server/zanata-model/src/main/java/org/zanata/model/Eraseable.java, line 42 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7TiPQ3Dw15YF-MxIt:-LC7gAiNCvIG8cqCueJC:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/zanata-model/src/main/java/org/zanata/model/Eraseable.java#L42)):*

It would be nice if we could avoid making all these mutator methods visible everywhere these entities are visible.
</blockquote></details>

Done.

---

*[server/zanata-model/src/main/java/org/zanata/model/Eraseable.java, line 44 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7SlUy7PtIal3KCafz:-LC7gBIlDOXSf1QHbqDR:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/zanata-model/src/main/java/org/zanata/model/Eraseable.java#L44)):*

I'm not entirely sure what this means without a javadoc, but if it erases the Erasable object, the name `erase` should be enough. 
</blockquote></details>

Done.

---

*[server/zanata-model/src/main/java/org/zanata/model/HAccount.java, line 212 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7V1jiBg_MnQIaQALY:-LC7gbpH4PuzAJVGI-u6:b-jm1omg) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/zanata-model/src/main/java/org/zanata/model/HAccount.java#L212)):*

I think we should try to make the erased usernames and erased email addresses similar. Maybe `_erased_{counter}`?

We shouldn't expose the primary key here. We need some sort of counter (or database sequence) for erasures.
</blockquote></details>

MySql don't have sequence. We could use timestamp or something arbitrary but I think the id is the most convenient. The username still appears in the UI (activity or translation history). I think we want to give the perception that this user is *deleted*. Although the fact that we just hide it.

---

*[server/zanata-model/src/main/java/org/zanata/model/HAccount.java, line 215 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7PXin5xZTyQUr7yyf:-LC7jC2Q0LirxNJnMiQW:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/zanata-model/src/main/java/org/zanata/model/HAccount.java#L215)):*

Perhaps we should erase the keys too: apiKey, accountActivationKey, accountResetPasswordKey

Maybe credentials, editorOptions, allowedApps too
</blockquote></details>

Done.

---

*[server/zanata-model/src/main/java/org/zanata/model/HPerson.java, line 177 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7U_J2COVjNKpFZK2w:-LC7jQLQ5aHLatdeHo83:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/zanata-model/src/main/java/org/zanata/model/HPerson.java#L177)):*

I think we should try to make the erased usernames and erased email addresses similar. Maybe `_erased_{counter}`?
</blockquote></details>

Done.

---

*[server/zanata-model/src/main/java/org/zanata/model/HPerson.java, line 177 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7VF8IEzpp-RFUYYK2:-LC7jRvs-88wd9DG4piV:b-nvfyr8) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/zanata-model/src/main/java/org/zanata/model/HPerson.java#L177)):*

We shouldn't expose the primary key here. We need some sort of counter (or database sequence) for erasures.
</blockquote></details>

no sequence in mysql

---

*[server/zanata-war/src/main/webapp-jboss/WEB-INF/classes/META-INF/persistence.xml, line 90 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7PouK8wo_Q4lUamcK:-LC7jV-J0LoV-YfL_4qo:b730e8h) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/zanata-war/src/main/webapp-jboss/WEB-INF/classes/META-INF/persistence.xml#L90)):*

Is this deliberate? Was this broken before?
</blockquote></details>

hard to set an environment variable with . in it. Maybe just me. But I think in general you don't want dot in shell variable.

---

*[server/zanata-war/src/main/webapp/admin/usermanager.xhtml, line 140 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7Q73P-Mha6qaE5Bmq:-LC7jo1e-jmQOOzKPazy:b-u7g5f6) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/zanata-war/src/main/webapp/admin/usermanager.xhtml#L140)):*

Perhaps we should hide this if the user has been erased.
</blockquote></details>

the user is erased and also disabled (so the UI will reflect that). We don't have UI for erased yet. It could come later with all UI parts and queries fixed.

---


*Comments from [Reviewable](https://reviewable.io/reviews/zanata/zanata-platform/811)*

seanf:
Reviewed 7 of 7 files at r3.
Review status: all files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io/reviews/zanata/zanata-platform/811)*

aeng:
Review status: all files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[server/services/src/main/java/org/zanata/service/impl/UserAccountServiceImpl.java, line 246 at r2](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7zPmr5w0f_2t7jXEO:-LCCUHVN8dxfh55DtjC2:b-i80fj0) ([raw file](https://github.com/zanata/zanata-platform/blob/c4e3d1393cb2f77971351c9c7a4dd24d2ddd0c28/server/services/src/main/java/org/zanata/service/impl/UserAccountServiceImpl.java#L246)):*

If we erase these HProjectMembers, we're going to end up with a lot of projects with no maintainer, without any sort of audit trail to explain why. Cleaning them up later will be very difficult. 

Perhaps we ought to keep the project/group memberships - are they really personal data? cc: @aeng
</blockquote></details>

According to the guideline, any information that is related to the person is considered personal data, and I think project falls into that category. Unless there is any use to keep the project/group membership, its best to remove them.

---


*Comments from [Reviewable](https://reviewable.io/reviews/zanata/zanata-platform/811)*

aeng:
Review status: all files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[server/services/src/main/java/org/zanata/service/impl/UserAccountServiceImpl.java, line 246 at r2](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7zPmr5w0f_2t7jXEO:-LCCUs_b40FDg2k8Gt3h:b-97viv5) ([raw file](https://github.com/zanata/zanata-platform/blob/c4e3d1393cb2f77971351c9c7a4dd24d2ddd0c28/server/services/src/main/java/org/zanata/service/impl/UserAccountServiceImpl.java#L246)):*
<details><summary><i>Previously, aeng (Alex Eng) wrote…</i></summary><blockquote>

According to the guideline, any information that is related to the person is considered personal data, and I think project falls into that category. Unless there is any use to keep the project/group membership, its best to remove them.
</blockquote></details>

I don't see any issue with cleaning up later, as I assume if it were implemented, we would just delete all projects without maintainer. 

---


*Comments from [Reviewable](https://reviewable.io/reviews/zanata/zanata-platform/811)*

aeng:
Review status: 20 of 21 files reviewed at latest revision, 9 unresolved discussions.

---

*[server/services/src/main/java/org/zanata/service/impl/UserAccountServiceImpl.java, line 232 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7WywEDtlCrEn1YE6Q:-LCCVjhD5cxoszQM0cof:b22ghae) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/services/src/main/java/org/zanata/service/impl/UserAccountServiceImpl.java#L232)):*

I understand, but if you leave them there, won't that lead to a constraint violation when you try to delete the project?

To put it another way, the handling of erased accounts' memberships seems inconsistent.  Shouldn't we delete them in all cases (barring a good reason not to, eg auditing), or else leave them all intact?
</blockquote></details>

Agreed with Sean. We should remove any memberships related to the project.

---


*Comments from [Reviewable](https://reviewable.io/reviews/zanata/zanata-platform/811)*

seanf:
Review status: 20 of 21 files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[server/services/src/main/java/org/zanata/service/impl/UserAccountServiceImpl.java, line 246 at r2](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7zPmr5w0f_2t7jXEO:-LCCXKJa2pb5uVkbvojh:bib6kj6) ([raw file](https://github.com/zanata/zanata-platform/blob/c4e3d1393cb2f77971351c9c7a4dd24d2ddd0c28/server/services/src/main/java/org/zanata/service/impl/UserAccountServiceImpl.java#L246)):*
<details><summary><i>Previously, aeng (Alex Eng) wrote…</i></summary><blockquote>

I don't see any issue with cleaning up later, as I assume if it were implemented, we would just delete all projects without maintainer. 
</blockquote></details>

I think there might already be a few with no maintainer (perhaps created by admin early on), but I guess that's fairly minor.

---

*[server/services/src/test/java/org/zanata/service/impl/UserAccountServiceImplTest.java, line 211 at r3](https://reviewable.io/reviews/zanata/zanata-platform/811#-LCCY-DFEsd6dD2PMjuR:-LCCY-DFEsd6dD2PMjuS:b-wf5rit) ([raw file](https://github.com/zanata/zanata-platform/blob/9b88f2dead50a8665dc6bb0ac96aa18e3318698c/server/services/src/test/java/org/zanata/service/impl/UserAccountServiceImplTest.java#L211)):*
> ```Java
>         assertThat(account.getAccountActivationKey()).isNull();
>         assertThat(account.getAccountResetPasswordKey()).isNull();
>         assertThat(account.getCredentials()).isEmpty();
> ```

Why remove this from the test?

---


*Comments from [Reviewable](https://reviewable.io/reviews/zanata/zanata-platform/811)*

huangp:
Review status: 20 of 21 files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[server/services/src/main/java/org/zanata/service/impl/UserAccountServiceImpl.java, line 195 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7x8IM8ALEqMlNo1Zu:-LCBkIl8CZ8bgP_QUJWU:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/services/src/main/java/org/zanata/service/impl/UserAccountServiceImpl.java#L195)):*

This method is very long. Try extracting parts into helper methods, eg

* deleteAccountObjects (keys, credentials, options)
* deleteReviewComments
* purgeSoleMaintainer
* softDeleteProjects
* deleteAccountMemberships
* deleteRawDocuments
</blockquote></details>

Done.

---

*[server/services/src/main/java/org/zanata/service/impl/UserAccountServiceImpl.java, line 213 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7XwO2ECqtIfIW6Wv3:-LCBkLV-ABoaharq30mS:b-z8g1rg) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/services/src/main/java/org/zanata/service/impl/UserAccountServiceImpl.java#L213)):*

Okay, didn't realise we were only soft deleting the projects, but given that HIterationGroup also supports soft delete, we should be consistent (I think we should soft-delete the group too).

(If an iteration group contains a lot of project versions, it may be very difficult to recreate.)

@aeng as we discussed the other day - I think purging/hiding project data (HProject, RawDocument, etc) from the translation memory is the wrong option for *corporate* instances of Zanata.  But I don't think it's urgent, so I'll create a separate issue.
</blockquote></details>

HIteration is not a soft delete. It's always a hard delete. 

---

*[server/services/src/main/java/org/zanata/service/impl/UserAccountServiceImpl.java, line 232 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7WywEDtlCrEn1YE6Q:-LCBkisN0ShNuYxbKBdF:boc9yhe) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/services/src/main/java/org/zanata/service/impl/UserAccountServiceImpl.java#L232)):*

I understand, but if you leave them there, won't that lead to a constraint violation when you try to delete the project?

To put it another way, the handling of erased accounts' memberships seems inconsistent.  Shouldn't we delete them in all cases (barring a good reason not to, eg auditing), or else leave them all intact?
</blockquote></details>

Right. Ok now I will only delete if the user is the *only* member in that project (regardless of its role). If there are someone else as language team coordinator, we only remove the membership and leave the project.

---

*[server/services/src/main/java/org/zanata/service/impl/UserAccountServiceImpl.java, line 276 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7WDf6CArWoe2mlTiO:-LCBmvRU-nR0ZtSg9av2:b-8diugm) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/services/src/main/java/org/zanata/service/impl/UserAccountServiceImpl.java#L276)):*

In some cases the user may have joined some teams but not created any documents or translations, which means the above deletions will allow the account to be deleted.
</blockquote></details>

Yes that may become possible. However to implement such thing is a bit tricky. You need either preemptively check whether such references still exist before deleting, or try it and catch the exception. If there is an exception, we may or may not be able to continue (think the hibernate not in threadsafe mode exception I showed you the other day). Again this can be something for later.

---

*[server/services/src/main/resources/messages.properties, line 1055 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7OCVjCtMf-d13RX7R:-LCBo-ov-PIKGY8Cu3rG:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/services/src/main/resources/messages.properties#L1055)):*

Typo: should be "user's"
</blockquote></details>

Done.

---

*[server/services/src/test/java/org/zanata/service/impl/UserAccountServiceImplTest.java, line 180 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7S4NO4ZAlvnjP_pUg:-LCCAil-5SGKVAgN6HFK:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/services/src/test/java/org/zanata/service/impl/UserAccountServiceImplTest.java#L180)):*

We should also test the other key/credentials tables here.
</blockquote></details>

Done.

---

*[server/services/src/test/java/org/zanata/service/impl/UserAccountServiceImplTest.java, line 211 at r3](https://reviewable.io/reviews/zanata/zanata-platform/811#-LCCY-DFEsd6dD2PMjuR:-LCCjoGe59queCnIal0L:b-igc5tl) ([raw file](https://github.com/zanata/zanata-platform/blob/9b88f2dead50a8665dc6bb0ac96aa18e3318698c/server/services/src/test/java/org/zanata/service/impl/UserAccountServiceImplTest.java#L211)):*

Why remove this from the test?
</blockquote></details>

Because the test data setup cause a ```java.lang.IllegalStateException: EntityManager is closed``` exception. It only happen when running this test with assignedFedoraRule together (see previous build or running it locally). I think it's to do with the UniqueValidator which unwraps a hibernate session from entityManager, also changing flush mode. Or it could also be the way we set up our test environment (transaction and CDI scopes etc are not real). I don't have the time to really dig into it... Feel free to tackle this if you have time. Maybe do it separately as this PR needs to be in production soon.

---

*[server/zanata-model/src/main/java/org/zanata/model/Eraseable.java, line 42 at r2](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7u62VEliuhTavDJ1G:-LCCEPlF9Cc5KEaKgO94:b-896fix) ([raw file](https://github.com/zanata/zanata-platform/blob/c4e3d1393cb2f77971351c9c7a4dd24d2ddd0c28/server/zanata-model/src/main/java/org/zanata/model/Eraseable.java#L42)):*

This is a mutator method too. Can we remove it from the interface?
</blockquote></details>

Done.

---

*[server/zanata-model/src/main/java/org/zanata/model/HAccount.java, line 184 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7TsQn81Ol7oNsWi7c:-LCCEQz-0f0FGht8_86L:b-plqnf1) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/zanata-model/src/main/java/org/zanata/model/HAccount.java#L184)):*

Please add these comments. The code in these methods looks a bit weird until you remember why we need this `new Date`.
</blockquote></details>

I don't think it's necessary. We have tons of place uses this. Unless you want to write a sed script to add comment to all of them. Findbugs will enforce this for us anyway.

---

*[server/zanata-model/src/main/java/org/zanata/model/HAccount.java, line 212 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7V1jiBg_MnQIaQALY:-LCCEkpg4P_yi1xh9-t9:b-gh12sc) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/zanata-model/src/main/java/org/zanata/model/HAccount.java#L212)):*

That's not enough reason to start exposing primary key IDs. 

Using a GUID might be another option.
</blockquote></details>

The reason I didn't choose UUID was: we have validation rules (not on entity) on username (size between 3-20). Although it's not enforced in database (255 varchar).  A GUID is 36 long. We can still store it. But I am not sure about what it might cause in other places (e.g. broken UI or something). I don't think exposing ID is that big a deal especially we are going to hide them from UI later on anyway. But if you insist I am fine to use GUID.

---


*Comments from [Reviewable](https://reviewable.io/reviews/zanata/zanata-platform/811)*

seanf:
Reviewed 3 of 3 files at r4.
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[server/services/src/main/java/org/zanata/service/impl/UserAccountServiceImpl.java, line 213 at r1](https://reviewable.io/reviews/zanata/zanata-platform/811#-LC7XwO2ECqtIfIW6Wv3:-LCRWbZ95vVWhcnGwtsh:biv7y6x) ([raw file](https://github.com/zanata/zanata-platform/blob/a626f6693574ae3bd034bb9da6ed427017bec6b8/server/services/src/main/java/org/zanata/service/impl/UserAccountServiceImpl.java#L213)):*
> H[Project]Iteration is not a soft delete. It's always a hard delete. [when deleted via the UI]

OK

---


*Comments from [Reviewable](https://reviewable.io/reviews/zanata/zanata-platform/811)*

seanf:
:lgtm:

---

---

*[server/services/src/test/java/org/zanata/service/impl/UserAccountServiceImplTest.java, line 211 at r3](https://reviewable.io/reviews/zanata/zanata-platform/811#-LCCY-DFEsd6dD2PMjuR:-LCWRGwZ2325yP-mvQj-:b14ds10) ([raw file](https://github.com/zanata/zanata-platform/blob/9b88f2dead50a8665dc6bb0ac96aa18e3318698c/server/services/src/test/java/org/zanata/service/impl/UserAccountServiceImplTest.java#L211)):*

Because the test data setup cause a ```java.lang.IllegalStateException: EntityManager is closed``` exception. It only happen when running this test with assignedFedoraRule together (see previous build or running it locally). I think it's to do with the UniqueValidator which unwraps a hibernate session from entityManager, also changing flush mode. Or it could also be the way we set up our test environment (transaction and CDI scopes etc are not real). I don't have the time to really dig into it... Feel free to tackle this if you have time. Maybe do it separately as this PR needs to be in production soon.
</blockquote></details>

Credentials assertion is back. 

OK

---


*Comments from [Reviewable](https://reviewable.io/reviews/zanata/zanata-platform/811#-:-LCWRVb43tynf-_ChzT_:bnfp4nl)*

spathare21:
Hi, 
Tested this PR and following are my observations :  

- Does user don't have any access to delete theire own data and his  account
- If we erased all data of user from  admin->manage users
-- Profile is not viewable after erasing users data 
-- Projects created by user is deleted 
-- Also user and projects is not searchable via explore page 
-- user unable to login if admin erase his all data
-- Erase users data deletes user and clean his all details 
-- It removes user from joined language teams


- Glossary created by user is not deleted 

screenshot from 2018-05-15 19-07-18


- It does not removes the user as project maintainer it shows hashing details 
screenshot from 2018-05-15 18-58-41


- Why do we have enable account option even after erasing all user data ?

screenshot from 2018-05-15 18-38-27


-  Unable to delete user account once we erase users personel data from manage users, is delete action removed ? if yes then please remove delete option 

screenshot from 2018-05-15 19-06-05


- Translation is not removed (in maintained project) which is done bye user whos all data is erased .


screenshot from 2018-05-15 19-08-07

@djansen-redhat this are the my observations not sure which are valid, but you can also have a look

seanf:
@spathare21 You mean project glossaries aren't deleted right? That is, glossaries for projects which are solely owned by the erased user?

I guess we should delete such glossaries. But I don't think it makes sense to delete global glossaries. Right @aeng ?

spathare21:
@seanf yup right

djansen-redhat:
I'm getting a nasty exception for the logged in user when they are deleted. Something must be left behind that identifies them, as it is using their original username.
https://pastebin.com/ETPghjzZ

djansen-redhat:
Maybe not a big deal for this PR, but 'Erased user' should not show up in Explore search results

djansen-redhat:
Do we need to prevent the kind of stupid where an admin can re-enable the account and set a password, allowing the user to log in?

djansen-redhat:
Prevent linking to user profile in textflow target history (it's a 404)

djansen-redhat:
`server/services/src/main/java/org/zanata/webtrans/server/rpc/GetTranslatorListHandler.java:63` can somehow be null, causing an error on the page (unseen?) and a null pointer exception on the server:
```
00:14:00,666 INFO  [org.zanata.webtrans.server.TranslationWorkspaceImpl] (default task-91) EditorClientId hc4fEewmMbA9Cd46aSw-lEOhXeL1QE3s7v65ceUl:6 has connectionId hc4fEewmMbA9Cd46aSw-lEOhXeL1QE3s7v65ceUl-5
00:14:01,328 ERROR [org.zanata.webtrans.server.SeamDispatch] (default task-90) Error dispatching action: java.lang.NullPointerException
	at org.zanata.webtrans.server.rpc.GetTranslatorListHandler.execute(GetTranslatorListHandler.java:63)
	at org.zanata.webtrans.server.rpc.GetTranslatorListHandler.execute(GetTranslatorListHandler.java:26)
	at org.zanata.webtrans.server.rpc.GetTranslatorListHandler$Proxy$_$$_WeldClientProxy.execute(Unknown Source)
	at org.zanata.webtrans.server.SeamDispatch.doExecute(SeamDispatch.java:167)
	at org.zanata.webtrans.server.SeamDispatch.execute(SeamDispatch.java:141)
	at org.zanata.webtrans.server.SeamDispatch$Proxy$_$$_WeldClientProxy.execute(Unknown Source)
	at org.zanata.webtrans.server.GwtDispatchService.execute(GwtDispatchService.java:39)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
```
Re-Indexing fixes this, so it may be related to index/cache, but maybe a null check is in order here.

huangp:
points from @spathare21 
Can not reproduce:
- erased user still appear in project maintainer list
- erased user still appear in explore page
Able to reproduce:
- erased user's translation still exist (this is expected. once they save their translation, it's considered a contribution to Red Hat and we will keep it)
- erased user's glossary (same as above. I think we don't need to remove them. Glossary terms are not related to personal data.
points from @djansen-redhat I think those UI related thing can come later. I also can't reproduce the NPE bug you mentioned.

djansen-redhat:
@huangp can confirm that the NPE for the http session occurs because the user was still logged in at the time.
Though a more unlikely case it's still a problem, but maybe can be fixed after this PR?

djansen-redhat:
@huangp this is my view of a project the user was a member of, not necessarily a maintainer:
screenshot from 2018-05-18 16-15-23

spathare21:
If you sort by role then also it will show delete users actions, if he is a member of project
screenshot from 2018-05-18 11-55-01

huangp:
fixed in last commit

spathare21:
@huangp 

- glossary created and translation done by user will not be removed right? 

- Also is Account enabled option is not fixed for the erased user? `Actually, I don't think so there is any meaning to keep account enabled options for deleted user

- If I have logged in user test123 and if admin erased his data then I am getting errors is it expected? 

screenshot from 2018-05-21 12-36-02

http://pastebin.test.redhat.com/593591

All other thing looks good to me

djansen-redhat:
@spathare21 the error you're seeing is the "delete user while they're logged in" problem. We don't have a fix for that yet. It will definitely need to be fixed before we introduce a "delete yourself" button, but it's bearable for now.

djansen-redhat:
I do believe this is good.

seanf:
:lgtm:

---

Reviewed 2 of 2 files at r5.

---



*Comments from [Reviewable](https://reviewable.io/reviews/zanata/zanata-platform/811#-:-LD4VS4l6FyiZHuQUDb7:bnfp4nl)*

PR #834, opened 2018-05-21
feat(ZNTA-2437) add option to get Approved translations only

dragosv:
JIRA issue URL: https://zanata.atlassian.net/browse/ZNTA-2437

Add option to get approved translations only

## Checklist

- JIRA link
- Check target branch

----


seanf:
@dragosv I haven't reviewed the tests yet (they're pretty hard to understand, but that's not a problem you introduced!). But I'll publish what I have reviewed so far.

---

Reviewed 22 of 32 files at r1.
Review status: 22 of 32 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---

*[api/zanata-common-api/src/main/java/org/zanata/common/MinContentState.java, line 27 at r1](https://reviewable.io/reviews/zanata/zanata-platform/834#-LDAPadR2OQbe4OORKp0:-LDAPadR2OQbe4OORKp1:b-puxfc5) ([raw file](https://github.com/zanata/zanata-platform/blob/172228e9363f69982a1d3e6e80072dd9948eb7ea/api/zanata-common-api/src/main/java/org/zanata/common/MinContentState.java#L27)):*
> ```Java
>  *         href="mailto:dvarovici.work@gmail.com">dvarovici.work@gmail.com</a>
>  */
> public enum MinContentState {
> ```

I don't think we should create a new enum for MinContentState. It's just a subset of ContentState. 

And I think it's more flexible if the client/api code just uses String, since it won't need to be recompiled or updated when the server adds a new ContentState.

If you don't want to delete MinContentState, I think a wrapper for String would have the flexibility, but still allow us to keep track of what type of String it represents.  Although I think ContentStateName would be a better name. And maybe move it out of zanata-common-api, since I don't think it's worth using such a wrapper for minContentState in the REST interfaces.

You could still define constants for Translated and Approved in ContentStateName, which means the usage shouldn't be very different.

---

*[api/zanata-common-api/src/main/java/org/zanata/rest/service/FileResource.java, line 260 at r1](https://reviewable.io/reviews/zanata/zanata-platform/834#-LD9jApA3RKakqNoJxHq:-LD9jApA3RKakqNoJxHr:bdaol30) ([raw file](https://github.com/zanata/zanata-platform/blob/172228e9363f69982a1d3e6e80072dd9948eb7ea/api/zanata-common-api/src/main/java/org/zanata/rest/service/FileResource.java#L260)):*
> ```Java
>                     @PathParam("fileType") String fileExtension,
>                     @QueryParam("docId") String docId,
>                     @QueryParam("minContentState") String minContentState);
> ```

I was going to suggest the enum `ContentState` here, but using a String is more flexible in terms of API evolution.

I think we should use String in the Resource interfaces (and in the corresponding implementation methods), convert to `ContentState` on the server as soon as possible and then use `ContentState` in all other server-side code.

As for the client, I think we should use String rather than an enum, and pass it through as a String, from CLI option right through to the REST calls. 

That way, if a new `ContentState` is added, let's say `FailedValidation` for example, the client doesn't need to be updated - the user just has to update the CLI option to say something like `--min-content-state=FailedValidation`.

---

*[api/zanata-common-api/src/main/java/org/zanata/rest/service/FileResource.java, line 260 at r1](https://reviewable.io/reviews/zanata/zanata-platform/834#-LDAS7AP0jwE7_qkW2RH:-LDAS7AP0jwE7_qkW2RI:b-j3tcj6) ([raw file](https://github.com/zanata/zanata-platform/blob/172228e9363f69982a1d3e6e80072dd9948eb7ea/api/zanata-common-api/src/main/java/org/zanata/rest/service/FileResource.java#L260)):*
> ```Java
>                     @PathParam("fileType") String fileExtension,
>                     @QueryParam("docId") String docId,
>                     @QueryParam("minContentState") String minContentState);
> ```

Please add `@DefaultValue("Translated")` - that way the `FileResourceImpl.downloadTranslationFile()` can assume the String `minContentState` is never `null` (even with old clients).

---

*[client/zanata-client-commands/src/main/java/org/zanata/client/commands/pull/PullOptionsImpl.java, line 228 at r1](https://reviewable.io/reviews/zanata/zanata-platform/834#-LDAQ2E-0kskEX9Y1zs8:-LDAQ2E-0kskEX9Y1zs9:brog9jl) ([raw file](https://github.com/zanata/zanata-platform/blob/172228e9363f69982a1d3e6e80072dd9948eb7ea/client/zanata-client-commands/src/main/java/org/zanata/client/commands/pull/PullOptionsImpl.java#L228)):*
> ```Java
>             usage = "Minimum content state to pull:\n"
>                     + "  \"translated\" (default) fetches approved or translated only.\n"
>                     + "  \"approved\" fetches approved only.\n")
> ```

Could you please capitalise Translated and Approved?

---

*[server/services/src/main/java/org/zanata/rest/service/FileService.java, line 243 at r1](https://reviewable.io/reviews/zanata/zanata-platform/834#-LDAUA1NACD_GxDQpsMr:-LDAUA1NACD_GxDQpsMs:bwiqwdj) ([raw file](https://github.com/zanata/zanata-platform/blob/172228e9363f69982a1d3e6e80072dd9948eb7ea/server/services/src/main/java/org/zanata/rest/service/FileService.java#L243)):*
> ```Java
> 
>             if (minContentState != null) {
>                 minContentStateObject = MinContentState.fromString(minContentState);
> ```

This could be done in one line with the ternary operator, but that won't be needed if minContentState gets a JAX-RS DefaultValue.

---


*Comments from [Reviewable](https://reviewable.io/reviews/zanata/zanata-platform/834#-:-LDAWbyp7WDoOJkv9W_v:b-g3gtsc)*

seanf:
Reviewed 2 of 2 files at r2, 3 of 4 files at r3, 1 of 1 files at r4.
Review status: 27 of 34 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[server/zanata-model-test/src/main/resources/org/zanata/test/model/ProjectsData.dbunit.xml, line 185 at r2](https://reviewable.io/reviews/zanata/zanata-platform/834#-LDAyJw_--2ykWWTkYI0:-LDAyJw_--2ykWWTkYI1:bdwx504) ([raw file](https://github.com/zanata/zanata-platform/blob/b5413a89fbe2447128115da4f51e6fd0e5a8d1f7/server/zanata-model-test/src/main/resources/org/zanata/test/model/ProjectsData.dbunit.xml#L185)):*
> ```XML
>   />
> 
>   <HProject defaultProjectType="File"
> ```

Could you please add an comment above this project data, explaining why it's here (eg which test uses it)?

---

*[server/zanata-war/src/test/java/org/zanata/rest/service/raw/MinContentStateRestITCase.java, line 18 at r3](https://reviewable.io/reviews/zanata/zanata-platform/834#-LDAzQv33rf8wOPQxyNj:-LDAzQv33rf8wOPQxyNk:b-a77icp) ([raw file](https://github.com/zanata/zanata-platform/blob/28bab984c50b62c890e00e5d5d8fae5b16297e8b/server/zanata-war/src/test/java/org/zanata/rest/service/raw/MinContentStateRestITCase.java#L18)):*
> ```Java
> public class MinContentStateRestITCase extends SourceAndTranslationResourceRestBase {
>     private static final String MULTIPLE_TEXT_FLOW_DATA_DB_UNIT_XML =
>             "org/zanata/test/model/MultipleTextFlowTestData.dbunit.xml";
> ```

Could I get you to rename that file to MinContentStateRestITCase.dbunit.xml? It might help to discourage the reuse of test data across different tests, which is a real test maintenance problem for us. 

Also, could you define a few constants here, similar to the ones I put in ProjectsData.kt, to replace the magic numbers in the test methods below?

eg APPROVED_TFTS, TRANSLATED_TFTS, ALL_TFTS

---

*[server/zanata-war/src/test/java/org/zanata/rest/service/raw/MinContentStateRestITCase.java, line 43 at r3](https://reviewable.io/reviews/zanata/zanata-platform/834#-LDB-IR7-YxgiOe9jRqr:-LDB-IR7-YxgiOe9jRqs:b-dlgpmi) ([raw file](https://github.com/zanata/zanata-platform/blob/28bab984c50b62c890e00e5d5d8fae5b16297e8b/server/zanata-war/src/test/java/org/zanata/rest/service/raw/MinContentStateRestITCase.java#L43)):*
> ```Java
>         TranslationsResource serverResource = getTranslationsResourceFromResponse(getResponse);
>         assertThat(serverResource.getTextFlowTargets()).hasSize(5);
>         assertThat(serverResource.getTextFlowTargets().stream().filter(textFlowTarget -> textFlowTarget.getState().equals(ContentState.Approved))).hasSize(2);
> ```

This would be more understandable if you use constants like `hasSize(APPROVED_TFTS + TRANSLATED_TFTS)`. See comment above.

---


*Comments from [Reviewable](https://reviewable.io/reviews/zanata/zanata-platform/834)*

seanf:
Thanks for the changes @dragosv. 

We don't use mock objects in the Arquillian tests in zanata-war, because all the tests in zanata-war share a single deployment (using a custom Arquillian setup).  

I'm concerned about the effect of installing MockTranslationMergeApproved as a CDI bean in zanata-war's tests, because it's going to affect all Arquillian tests which use TranslationMergeService, not just whichever test needs it right now - which test is that, by the way? Or multiple tests? 

If I can understand what's going on, I might be able to adapt the relevant test to work in the services module, using an Arquillian "micro-deployment" (see AccountRawRestITCase for an example).

Could you please explain (broadly) what you changed in each of AsyncResourceRestITCase, ResourceServiceRestITCase and FileRawRestITCase? Looks like you did some good refactoring and de-duplication, but it's so different I can't tell what's refactoring and what's actually different.

---

Review status: 33 of 41 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[client/zanata-client-commands/src/test/java/org/zanata/client/commands/pull/PullCommandTest.java, line 232 at r9](https://reviewable.io/reviews/zanata/zanata-platform/834#-LDeL2kd-BNWTBIKzCw5:-LDeL2kd-BNWTBIKzCw6:bm43uaz) ([raw file](https://github.com/zanata/zanata-platform/blob/3c1eac7907388532b78665089f37b8462a32b31c/client/zanata-client-commands/src/test/java/org/zanata/client/commands/pull/PullCommandTest.java#L232)):*
> ```Java
>     }
> 
>     private PullCommand getPullCommand() {
> ```

Thanks for cleaning that up!

---

*[server/zanata-war/src/test/java/org/zanata/rest/service/raw/MinContentStateRestITCase.java, line 57 at r4](https://reviewable.io/reviews/zanata/zanata-platform/834#-LDG7Kx6-qtpDFxNfw5j:-LDG7Kx6-qtpDFxNfw5k:bl9hiiv) ([raw file](https://github.com/zanata/zanata-platform/blob/a4bdfadb4e5419be94c5588dc79b4765e5ac0fc5/server/zanata-war/src/test/java/org/zanata/rest/service/raw/MinContentStateRestITCase.java#L57)):*
> ```Java
>         assertThat(serverResource.getTextFlowTargets()).hasSize(5);
>         assertThat(serverResource.getTextFlowTargets().stream().filter(textFlowTarget -> textFlowTarget.getState().equals(ContentState.Approved))).hasSize(1);
>         assertThat(serverResource.getTextFlowTargets().stream().filter(textFlowTarget -> textFlowTarget.getState().equals(ContentState.Translated))).hasSize(1);
> ```

It would be good to replace these numbers with constants too. Thanks.

---

*[api/zanata-common-api/src/main/java/org/zanata/common/MinContentState.java, line 27 at r1](https://reviewable.io/reviews/zanata/zanata-platform/834#-LDAPadR2OQbe4OORKp0:-LDecfeI5_0Pi4zZ1v23:b-gcor2j) ([raw file](https://github.com/zanata/zanata-platform/blob/172228e9363f69982a1d3e6e80072dd9948eb7ea/api/zanata-common-api/src/main/java/org/zanata/common/MinContentState.java#L27)):*

I don't think we should create a new enum for MinContentState. It's just a subset of ContentState. 

And I think it's more flexible if the client/api code just uses String, since it won't need to be recompiled or updated when the server adds a new ContentState.

If you don't want to delete MinContentState, I think a wrapper for String would have the flexibility, but still allow us to keep track of what type of String it represents.  Although I think ContentStateName would be a better name. And maybe move it out of zanata-common-api, since I don't think it's worth using such a wrapper for minContentState in the REST interfaces.

You could still define constants for Translated and Approved in ContentStateName, which means the usage shouldn't be very different.
</blockquote></details>

I meant that ContentStateName could be a wrapper class containing a string, to be used on the client, not an enum to use on the server. I've added a commit to remove ContentStateName and just use ContentState on the server.

---


*Comments from [Reviewable](https://reviewable.io/reviews/zanata/zanata-platform/834#-:-LDedi9y9qvFsZrIFr_g:b-j95zif)*

seanf:
@dragosv Did you do some sort of weird merge or rebase recently? Several changes from master are showing up as new commits ("authored" by other devs, "committed" by you), and had to be reviewed again.

seanf:
@dragosv Please disregard the repetition of "It would be good to replace these numbers with constants too. Thanks." - not sure why that reappeared.

I'd still like to know the reasons for the new data in ProjectsData.dbunit.xml, though. Thanks!

seanf:
@dragosv Looks like it's just FileRawRestITCase.getBakedGetTextApprovedImportAsApproved that needs MockTranslationMergeApproved. I'll see what I can do with it.

---

Review status: 33 of 41 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io/reviews/zanata/zanata-platform/834#-:-LDeqmdV-MyQjQZKD988:b-q1bgox)*

seanf:
@dragosv, Digging into the code a bit more for your review, I think some of my implementation advice was bad, so I'm trying to change a number of things, fairly drastically I'm afraid. I'm sorry about breaking your branch and some tests in the meantime.

But please don't change any of the code until I get things back into shape. It might take me a few days. And don't worry about those Mock classes.

@dragosv If you specify the command line option `--min-content-state=Approved` (or `--approved`, as I may change it to), what would you expect to receive on the client for strings which are merely Translated?

- In Properties files? (omit the property, I guess?)
- In Gettext files? (marked as Fuzzy perhaps? Or omitted?)
- In HTML files? (replace with a blank string? or the original language text?)
- In Zanata "xml" files? (Are you using this semi-undocumented format?)

In your use case, what types of files are you sending in and out of Zanata anyway?

Please let me know.

dragosv:
@seanf no problem changing the code

In regards to the command line option, when specifying or --approved (prefer this over --min-content-state=Approved), this is what I want for a proper integration with Serge when the item is only Translated but not approved

In Properties files: Omit
In Gettext files: Omit 
In HTML files: Whatever makes sense (probably makes sense to put the original)

Xliff
Xliff 1.1 (see https://zanata.atlassian.net/browse/ZNTA-2562): Omit 
Xliff 1.2 (I will implement, see above): Use xliff state (signed-off for approved, translated for translated)
Xliff 2.0 (Future plan to implement): Use xliff state (final for approved, translated for translated)

In Zanata "xml" files? : Do not know

I will be using only gettext or xliff 1.2/ 2.0 for communicating with Serge

seanf:
> In regards to the command line option, when specifying or --approved (prefer this over --min-content-state=Approved), this is what I want for a proper integration with Serge when the item is only Translated but not Approved

I decided to call it approvedOnly everywhere, with the CLI option --approved-only. We could probably alias it as -a or something if necessary.

I still want to apply the approvedOnly change to FileResource/Service, but it should be straightforward enough now. It's unfortunate about the overlap with "half-baked" and "baked", but I don't want to add to that mess by creating "fully-baked" and "po-fully-baked"!

The problem with minContentState is that there's no **real** concept of minimum. The ordinal values for ContentState just aren't in a meaningful order - Approved was added after Translated, so its ordinal is higher. Besides, which is better: NeedReview (Fuzzy) or Rejected? I'm not sure. So yeah, my minContentState suggestion was a bad one. Sorry! Thus approvedOnly (and markTranslatedAsApproved in TranslatedDocResource - since that's what the old implementation was **really** doing).

> In Properties files: Omit

Yep, unless you use the createSkeletons option, which puts in an empty string.

> In Gettext files: Omit (Maybe Fuzzy, ideally should be an option)

I decided to include them in the PO file, but marked as Fuzzy. It doesn't make sense to keep NeedReview items as Fuzzy, but not keep Translated items. But when you parse the PO file, just ignore fuzzy messages (which you already needed to do) and you'll get the result you want.

I guess it could sometimes make sense to omit both NeedReview and Translated when --approved-only is set, but I probably won't have time to implement that myself. There's actually an includeFuzzy option which is currently incompatible with --approved-only - I guess it could be used here.

> In HTML files: Whatever makes sense (probably makes sense to put the original)

Yep.

> - Xliff
>  - Xliff 1.1 (see https://zanata.atlassian.net/browse/ZNTA-2562): Omit
>  - Xliff 1.2 (I will implement, see above): Use xliff state (signed-off for approved, translated for translated) or Omit

I wasn't _quite_ sure if you meant to implement it yourself. Great, thanks!

>  - Xliff 2.0 (Future plan to implement): Use xliff state (final for approved, translated for translated) or Omit

Yep, that all makes sense.

> In Zanata "xml" files? : Do not know

That's fine. They're just meant to be an exact copy of what came back from the server anyway.

> I will be using only gettext or xliff 1.2/ 2.0 for communicating with Serge. By default Serge has gettext support only, however I am developing xliff 1.2 and 2.0 (Xliff 1.2 is now a pull request evernote/serge#67). The gettext format considers whatever it receives as translated

Except fuzzy, I hope.

> while the future xliff 1.2 has support to filter based on xliff state.

@dragosv I take it you are doing everything with "File" projects in Zanata (not Gettext, podir or XLIFF projects)?

dragosv:
@seanf let me know what changes are you stiill expecting from me. 

Re MockTranslationMergeApproved i needed it so that when translations are merged they are approved instead of moved into translated. However after your changes I do not see it used any more.

In regards to AsyncResourceRestITCase, ResourceServiceRestITCase and FileRawRestITCase it is refactoring for the purpose of removing duplication and also conforming with SourceAndTranslationResourceRestBase based tests (the good parts). I will try to explain it more if you want. How should I do that? Is there a way to code review/comment in one of the tools used?

seanf:
@dragosv I'm not really expecting any changes from you right now. I've been trying to squash your commits and mine (separately) so that the git history won't be too messy, then I'll get our QA team to take a look. But apparently I broke something, so I've got to get that working.

MockTranslationMergeApproved was never used by name, but it's currently needed by FileRawRestITCase. After this PR, I plan to create a ContentState parameter for  TranslatedDocResource.putTranslationsWithDocId and  FileResource.uploadTranslationFile, so that you can upload translations as Approved (depending on user permissions). This should remove the need for MockTranslationMergeApproved, and fill a functionality gap. I don't want that to hold up this PR though.

You can reply to individual comments in Reviewable: https://reviewable.io/reviews/zanata/zanata-platform/834 - there's a toolbar across the top which make it easy to find the next comment to reply to.

You're just using Zanata's File projects, is that right?

seanf:
@dragosv If you're responding on https://reviewable.io/reviews/zanata/zanata-platform/834 , please note that you have to hit the Publish button (top right) before I can see your comments.

seanf:
@dragosv Looks like that build failure was just a problem with a Jenkins plugin.

I see you added another test downloadXliff11TranslationApprovedOnly, but I think you forgot to set approvedOnly to true.

seanf:
(cc @djansen-redhat)

@dragosv When I asked if you were using File projects, I was referring to this setting: http://docs.zanata.org/en/release/images/create-project-completed.png. It's possible to use Gettext with "File" projects and with "Gettext" or "Podir" projects, although "File" tends to be the better option these days.

I finished my refactoring from minContentState to approvedOnly, but then I added some more tests and found that the Okapi adapters don't even handle NeedReview/Fuzzy properly, never mind approvedOnly. 

Plus there seems to be a problem with uploading "offline PO" gettext against Okapi documents in a File project, _and_ the TS adapter seems to mark translations as "unfinished" whether Approved, Translated or NeedReview.

I think the "offline PO" problem can be tackled separately, but I think the NeedReview and approvedOnly problems are pretty much inseparable. I'm working on it.

Are there more changes you want to make?

dragosv:
@seanf No more changes as far I am concerned. On another note both made changes to the same test  :)

dragosv:
@seanf There are still a bunch of codefactor and codebeat issues, shouldn't those be fixed?

seanf:
> In regards to the File project I have not played with it.

So are your projects configured as "podir" then?

> Can you elaborate why the File project is the better option? Also should this project type be preferred with xliff files as well?

File projects are the only ones which can use Okapi adapters, plus they can support a variety of different file types within the same project. So when you want to use your future XLIFF 1.2 support, then you will have to use File projects. For the current XLIFF 1.1 support, an "XLIFF" project is fine.

> However, there are still a bunch of codefactor and codebeat issues, shouldn't those be fixed?

Well, ideally they should, but I wasn't going to make you fix all the pre-existing problems with the codebase!

The code coverage (**codecov/patch**) is a bit of a concern, but our coverage stats only cover unit tests, so the tests in FileRawRestITCase, while important, aren't counted. Also, the codecov reports are often incomprehensible because they have missing information.

Codacy and CodeFactor don't seem to have found anything that is both serious and new.

Codebeat is complaining about a lot of methods which have even more parameters than before (eg approvedOnly). Have you got any good ideas? Honestly I'm a bit worried about big refactoring for some of these because they may not have a lot of automated test coverage, but I'm open to suggestions.

dragosv:
I am using gettext projects for now. In future I plan on using file + xliff 1.2 / xliff 2.0.
No good idea regards to more parameters than before issue.

When can I expect this to be QA? Asking as I would like to use the approved only in serge.

seanf:
Note to QA: this PR affects the code for downloading translations: via CLI or Maven plugin, both for File projects and non-File projects (eg Properties or podir). It adds a new option `--approved-only` (`approvedOnly` in Maven) which will exclude merely Translated strings from the downloaded document.

Note that there's no approvedOnly option for downloads via web UI at this stage.

seanf:
@dragosv I think we can tackle the too-many-parameters problem by introducing some new parameter objects:

```kotlin
// This one will really be a Java class:
data class TranslatedDoc { // or ResourceWithTranslation? in zanata-common-api
// To be used pretty much everywhere we have a Resource with TranslationResource.
	val source: Resource
	val translation: TranslationResource
	val locale: LocaleId
}

// in org.zanata.adapter.FileFormatAdapter: parseDocumentFile(), parseTranslationFile()
data class FileFormatAdapter.ParserOptions { // in services
    val rawFile: URI // (formerly originalFile/originalDoc)
    val params: Optional<String>
    val locale: LocaleId
}

// in org.zanata.adapter.FileFormatAdapter.writeTranslatedFile()
data class FileFormatAdapter.WriterOptions { // in services
	val parserOptions: ParserOptions
	val translatedDoc: TranslatedDoc
}
```

But we won't let it hold up this PR.

I'm hoping this can go through QA in the next few days, and merged into `master` very soon after, but we probably won't be releasing Zanata 4.6 for a couple of months. We're just doing the release testing for 4.5 now.

djansen-redhat:
@dragosv @seanf I'm pretty confident this does the job. Tested on several project and file types.

seanf:
Reviewed 1 of 32 files at r1, 1 of 24 files at r5, 1 of 7 files at r10, 63 of 74 files at r11, 3 of 4 files at r12, 1 of 2 files at r13, 1 of 1 files at r14, 22 of 22 files at r15.
Review status: :shipit: complete! all files reviewed, all discussions resolved

---



*Comments from [Reviewable](https://reviewable.io/reviews/zanata/zanata-platform/834)*

PR #849, opened 2018-05-25
fix(ZNTA-2594): Improve handling of raw translation document upload

djansen-redhat:
JIRA issue URL: https://zanata.atlassian.net/browse/ZNTA-1083

If .po, then process as a .po, otherwise if it's a File project then try to process it with "good intentions".

## Checklist

- JIRA link
- Check target branch

----


djansen-redhat:
**A bit of a UX improvement. Tell the user that anything other than a .po is "YMMV"...**
screenshot from 2018-05-26 12-47-28
**.pot update is still as normal...**
screenshot from 2018-05-26 12-48-55
**Last warning!**
screenshot from 2018-05-26 12-47-56

djansen-redhat:
@aeng @seanf what do you think of this?

aeng:
@djansen-redhat I think this change makes sense on how the platform should manage the po/pot vs file project type.

aeng:

---



*Comments from [Reviewable](https://reviewable.io/reviews/zanata/zanata-platform/849)*

aeng:

---



*Comments from [Reviewable](https://reviewable.io/reviews/zanata/zanata-platform/849)*

djansen-redhat:
@seanf I fixed the issue with the translationExtensionOf null pointer. Gettext and Null projects show .po, all others show .po, $extension.

seanf:
Review status: 4 of 5 files reviewed at latest revision, all discussions resolved.

---

*[server/services/src/main/resources/messages.properties, line 634 at r4](https://reviewable.io/reviews/zanata/zanata-platform/849#-LDeG-AM9ucm7iRw8XVt:-LDeG-AM9ucm7iRw8XVu:bje9mod) ([raw file](https://github.com/zanata/zanata-platform/blob/d67248373cb3e4e043668c03615e5e9ebd5e4e48/server/services/src/main/resources/messages.properties#L634)):*
> ```INI
> jsf.Upload.Label=Upload translation (.po{0})
> jsf.Upload.Description=Translation file (.po) upload
> jsf.Upload.DescriptionFile=Translation file (.po) upload. Upload of a translated document in original format to a File project type is experimental.
> ```

I would suggest wording something like this: "Upload of a translated document in original format is experimental and translations may not be recognised depending on the file format."

---

*[server/zanata-war/src/main/webapp/WEB-INF/layout/version/languages-tab.xhtml, line 397 at r4](https://reviewable.io/reviews/zanata/zanata-platform/849#-LDeI15998lD_ssxN_Nd:-LDeI15998lD_ssxN_Ne:bvewftp) ([raw file](https://github.com/zanata/zanata-platform/blob/d67248373cb3e4e043668c03615e5e9ebd5e4e48/server/zanata-war/src/main/webapp/WEB-INF/layout/version/languages-tab.xhtml#L397)):*
> ```HTML
>                         </li>
>                         <ui:fragment
>                             rendered="#{not versionHomeAction.isPoDocument(document.docId) and versionHomeAction.isFileUploadAllowed(versionHomeAction.selectedLocale) and versionHomeAction.isUserAllowedToTranslateOrReview(versionHomeAction.selectedLocale)}">
> ```

All these 'rendered' conditionals are too complicated to understand. We should move them into property methods with meaningful names in VersionHomeAction, eg `rendered="#{versionHomeAction.rawFileUploadAllowed}"` which calls `isRawFileUploadAllowed()`.

---


*Comments from [Reviewable](https://reviewable.io/reviews/zanata/zanata-platform/849)*

djansen-redhat:
Review status: 2 of 10 files reviewed at latest revision, 1 unresolved discussion.

---

*[server/services/src/main/resources/messages.properties, line 634 at r4](https://reviewable.io/reviews/zanata/zanata-platform/849#-LDeG-AM9ucm7iRw8XVt:-LDg1NlT3b49Px5aPotN:b-rqvks2) ([raw file](https://github.com/zanata/zanata-platform/blob/d67248373cb3e4e043668c03615e5e9ebd5e4e48/server/services/src/main/resources/messages.properties#L634)):*

I would suggest wording something like this: "Upload of a translated document in original format is experimental and translations may not be recognised depending on the file format."
</blockquote></details>

I uh... sorta did this. Probably need to discuss the wording.

---

*[server/zanata-war/src/main/webapp/WEB-INF/layout/version/languages-tab.xhtml, line 397 at r4](https://reviewable.io/reviews/zanata/zanata-platform/849#-LDeI15998lD_ssxN_Nd:-LDg1V8R4yuXSUkrsXyZ:b-vqdm9h) ([raw file](https://github.com/zanata/zanata-platform/blob/d67248373cb3e4e043668c03615e5e9ebd5e4e48/server/zanata-war/src/main/webapp/WEB-INF/layout/version/languages-tab.xhtml#L397)):*

All these 'rendered' conditionals are too complicated to understand. We should move them into property methods with meaningful names in VersionHomeAction, eg `rendered="#{versionHomeAction.rawFileUploadAllowed}"` which calls `isRawFileUploadAllowed()`.
</blockquote></details>

Done, and added an adapter condition for "can upload raw".

---


*Comments from [Reviewable](https://reviewable.io/reviews/zanata/zanata-platform/849)*

seanf:
Review status: 6 of 10 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---

*[server/services/src/main/java/org/zanata/action/VersionHomeAction.java, line 1341 at r5](https://reviewable.io/reviews/zanata/zanata-platform/849#-LDitT2J3oTjqm5F9OB1:-LDitT2J3oTjqm5F9OB2:b3o4uxo) ([raw file](https://github.com/zanata/zanata-platform/blob/9b0b34cbed18ffe731962fd9e37e8be495d9583b/server/services/src/main/java/org/zanata/action/VersionHomeAction.java#L1341)):*
> ```Java
>         boolean canUpload = false;
>         if (!hasOriginal(docPath, docName)) {
>             return false;
> ```

Technically, you don't really need the original file to be on the file system, because we only parse the translation file. Although if you've lost the source file, there's probably no point in uploading translations (unless someone uploads the source file again).

However, we could actually make translation uploads for index-based adapters (using IdSource.textUnitId) a bit safer by parsing source and translation documents in parallel, rejecting any documents where the Okapi events don't match (ie different document structure). Fortunately, we don't have any adapters which use IdSource.textUnitId right now, since they're generally a bad idea.

---

*[server/services/src/main/java/org/zanata/adapter/FileFormatAdapter.java, line 149 at r5](https://reviewable.io/reviews/zanata/zanata-platform/849#-LDiw7Mx3IkLkR8e63MF:-LDiw7Mx3IkLkR8e63MG:b65dmqu) ([raw file](https://github.com/zanata/zanata-platform/blob/9b0b34cbed18ffe731962fd9e37e8be495d9583b/server/services/src/main/java/org/zanata/adapter/FileFormatAdapter.java#L149)):*
> ```Java
>     default boolean rawTranslationUploadAvailable() {
>         return false;
>     }
> ```

I think it would be good to have an implementation in `OkapiFilterAdapter.rawTranslationUploadAvailable()` which will return `idSource == textUnitName`. Any Okapi filter which uses another IdSource is probably not going to work, and any filter which does use textUnitName should be okay. So all those subclasses won't need to implement `rawTranslationUploadAvailable()` individually.

I think we should return `true` in `GettextAdapter.rawTranslationUploadAvailable()`. Properties might work, but I don't think it's a good idea (since the source text is not included), so `false` is probably right.

XLIFF does include the source text, but I don't think it will work with our current implementation, so `false` is probably good there too.

---

*[server/services/src/main/java/org/zanata/adapter/TSAdapter.java, line 377 at r5](https://reviewable.io/reviews/zanata/zanata-platform/849#-LDivuJVAr_AxmTXOHu4:-LDivuJVAr_AxmTXOHu5:bwqmoia) ([raw file](https://github.com/zanata/zanata-platform/blob/9b0b34cbed18ffe731962fd9e37e8be495d9583b/server/services/src/main/java/org/zanata/adapter/TSAdapter.java#L377)):*
> ```Java
>     @Override
>     public boolean rawTranslationUploadAvailable() {
>         return true;
> ```

This adapter uses contentHash - it doesn't seem like uploading translation files will work very well.

---


*Comments from [Reviewable](https://reviewable.io/reviews/zanata/zanata-platform/849)*

djansen-redhat:
TODO:
- [ ] when #858 is merged, rebase
- [ ] enable + test properties upload
- [ ] add logic in UI to show when a raw format translation upload is available
- [ ] add logic in adapter code to indicate it is capable of raw format translation parsing

seanf:
I think this PR needs a new name and a new JIRA, since https://github.com/zanata/zanata-platform/pull/858 seems to cover https://zanata.atlassian.net/browse/ZNTA-1083

djansen-redhat:
Most of this got swallowed up by #854

PR #854, opened 2018-05-29
fix(ZNTA-2581,ZNTA-1731): fix textflow id of uploaded JSON documents

djansen-redhat:
JIRA issue URL: https://zanata.atlassian.net/browse/ZNTA-2581

Use the full path of a JSON object to determine its id. This disallows standalone/isolated strings to be detected as the path names will conflict when there are multiple entities, such as an array of strings.

- [x] Need to handle duplicate IDs, eg a key "text/2" vs a nested key "2" inside "text" may both map to "/text/2"

## Checklist

- JIRA link
- Check target branch

----


seanf:
@djansen-redhat I've added a one-item checklist to make sure we don't forget about the duplicate/nesting problem.

djansen-redhat:
@spathare21 -
- upload a JSON source file
- upload a translated version of the JSON file as a translation (last option in the drop down, should have a * and a warning tooltip about RAW formats)
- download the result

It should show the source file now translated.

PR #925, opened 2018-06-23
fix(ZNTA-2634): replace language property in translated TS

djansen-redhat:
JIRA issue URL: https://zanata.atlassian.net/browse/ZNTA-2634

Use a Regex to replace the language="$" property in a Qt TS DocumentPart.

## Checklist

- JIRA link
- Check target branch

----


djansen-redhat:
@seanf is this _too_ hacky, you think?

server/services/src/main/java/org/zanata/adapter/TSAdapter.java
```diff
@@ -252,6 +256,19 @@ private boolean usable(ContentState state, boolean approvedOnly) {
...
+    /**
+     * Replace the language="en_US" with a target localeId
+     * @param event the DocumentPart event of a TS file header
+     * @param localeId the desired target locale
+     * @return a GenericSkeleton with the language replaced
+     */
+    private ISkeleton replaceLocaleInDocPart(Event event, String localeId) {
+        DocumentPart part = event.getDocumentPart();
+        return new GenericSkeleton(part.getSkeleton().clone().toString()
+                .replaceAll("language\\s*=\\s*\".+\"",
```
seanf: This regex could mess with the sourcelanguage attribute. Also, it would be a bit safer to use `[^"]` instead of `.` between the quotes.

http://doc.qt.io/archives/qt-4.8/linguist-ts-file-format.html

server/services/src/main/java/org/zanata/adapter/TSAdapter.java
```diff
@@ -196,7 +199,8 @@ protected void generateTranslatedFile(URI originalFile,
                 Event event = filter.next();
                 if (event.isDocumentPart() &&
                         event.getDocumentPart().hasProperty("language")) {
-                    // TODO ZNTA-2483 change readonly language property
+                    event.getDocumentPart().setSkeleton(
+                            replaceLocaleInDocPart(event, localeId.toString()));
```
seanf: toBCP47 would be a bit safer, in fact why don't you pass the Okapi LocaleId into the method? That way it won't have to check the string against a regex.

server/services/src/main/java/org/zanata/adapter/TSAdapter.java
```diff
@@ -252,6 +256,35 @@ private boolean usable(ContentState state, boolean approvedOnly) {
...
+        try {
+            part = event.getDocumentPart().getSkeleton().clone().toString();
+        } catch (OkapiUnexpectedResourceTypeException okapiException) {
+            log.error("Unexpected Qt TS event type: {}, in {}",
+                    event.getEventType(), event.getResource().getSkeleton());
+            throw new FileFormatAdapterException("Qt TS Adapter error");
+        }
+
+        try {
+            part = part.replaceAll("\\slanguage\\s*=\\s*\"[\\w\\d@.-]*\"",
```
seanf: This regex would be easier to read as a raw string in Kotlin.

server/services/src/test/java/org/zanata/adapter/TSAdapterTest.java
```diff
@@ -254,6 +258,74 @@ public void testFailToParseOriginalFile() throws Exception {
...
+        assertDocpartStringIsChanged("mo", true);
+        assertDocpartStringIsChanged("mo_PH", true);
+        assertDocpartStringIsChanged("mo-PH", true);
+        assertDocpartStringIsChanged("mo_PH.UTF-8", true);
+        assertDocpartStringIsChanged("mo_PH.UTF-8@5", true);
+    }
+
+    private void assertDocpartStringIsChanged(String originalLoc,
+                                              boolean changed) {
+        String base = "<TS version=\"2.1\" language=\"REPLACE\">";
```
seanf: To protect against future breakage, could you please add another test case using the sourcelanguage attribute (which shouldn't be modified)?

spathare21:
@djansen-redhat  looks fine language property is showing correctly now as target lang 
<TS version="2.1" language="de">

djansen-redhat:
Verified by @spathare21

PR #932, opened 2018-06-26
fix(test): convert admin tests to Kotlin and fix

djansen-redhat:
JIRA issue URL: n/a

Kotlin-fy the admin functional tests

## Checklist

- JIRA link
- Check target branch

----


server/functional-test/src/test/java/org/zanata/feature/administration/AutoRoleAssignmentTest.kt
```diff
@@ -18,55 +18,54 @@
...
+@Category(DetailedTest::class)
+class AutoRoleAssignmentTest : ZanataTestCase() {
 
     @Trace(summary = "The administrator can create a rule to assign roles " +
             "at user sign in")
-    @Test(timeout = ZanataTestCase.MAX_SHORT_TEST_DURATION)
-    public void createAutoRoleAssignment() throws Exception {
-        RoleAssignmentsPage roleAssignmentsPage = new LoginWorkFlow()
+    @Test(timeout = MAX_SHORT_TEST_DURATION.toLong())
+    @Throws(Exception::class)
```
seanf: Not a problem, but feel free to delete these Throws annotations on tests.

server/functional-test/src/test/java/org/zanata/feature/administration/AutoRoleAssignmentTest.kt
```diff
@@ -18,55 +18,54 @@
...
                 .enterIdentityPattern(".+ransla.+")
                 .selectRole("admin")
-                .saveRoleAssignment();
+                .saveRoleAssignment()
 
-        assertThat(roleAssignmentsPage.getRulesByPattern())
-                .as("The rule was created")
-                .contains(".+ransla.+");
+        assertThat(roleAssignmentsPage.rulesByPattern)
+                .`as`("The rule was created")
```
seanf: This can be nicer in Kotlin with the name `describedAs` (no need for back-ticks).

server/functional-test/src/test/java/org/zanata/feature/administration/ManageSearchTest.kt
```diff
@@ -18,89 +18,93 @@
...
             "search indexes")
-    @Test(timeout = ZanataTestCase.MAX_SHORT_TEST_DURATION)
+    @Test(timeout = ZanataTestCase.MAX_SHORT_TEST_DURATION.toLong())
     @Ignore("Unstable - sometimes the button isn't ready and sometimes " +
             "the index fails to complete.")
-    public void regenerateSearchIndexes() throws Exception {
-        ManageSearchPage manageSearchPage = dashboardPage
+    @Throws(Exception::class)
+    fun regenerateSearchIndexes() {
+        var manageSearchPage = dashboardPage!!
```
seanf: Instead of using the "not null assertion", it's often better to use a non-nullable type with `lateinit` (eg `private lateinit var dashboardPage: DashboardBasePage`), since it does get set to a non-null value before use, in this case by the JUnit Before method.

djansen-redhat:
Part of https://github.com/zanata/zanata-platform/pull/946

PR #935, opened 2018-06-27
fix(test): convert concurrentedit tests to Kotlin

djansen-redhat:
JIRA issue URL: n/a

Convert the concurrent editing functional tests to Kotlin

## Checklist

- JIRA link
- Check target branch

----


server/functional-test/src/test/java/org/zanata/feature/concurrentedit/ConcurrentAccessTest.kt
```diff
@@ -0,0 +1,129 @@
...
+                iterationSlug, "gettext")
+
+        val threadCount = 5
+        val task = Callable {
+            val suffix = counter.getAndIncrement()
+            ZanataRestCaller().postSourceDocResource(
+                    projectSlug, iterationSlug, buildResource(suffix),
+                    false)
+        }
+        val tasks = Collections.nCopies(threadCount, task)
```
seanf: If you inline task as a lambda, you won't need to declare it as Callable.

server/functional-test/src/test/java/org/zanata/feature/concurrentedit/ConcurrentAccessTest.kt
```diff
@@ -0,0 +1,129 @@
...
+
+        private fun buildResource(suffix: Int): Resource {
+            return buildSourceResource("doc$suffix",
+                    buildTextFlow("res$suffix", "content$suffix"))
+        }
+
+        private fun getStatusCodes(futures: List<Future<Int>>): MutableList<Int> {
+            return futures.stream()
+                    .map { it -> getResult(it) }
+                    .collect(Collectors.toList())
```
seanf: If this were new Kotlin code, I would write it like this:

```kt
        private fun getStatusCodes(futures: List<Future<Int>>): MutableList<Int> =
            futures.map { getResult(it) }.toMutableList()
```

Then there should be a quick-fix to use a method reference, probably something like this:

```kt
            futures.map { this::getResult) }.toMutableList()
```

PR #941, opened 2018-06-29
fix(test): convert editor tests to Kotlin

djansen-redhat:
JIRA issue URL: n/a

Convert the editor functional tests to Kotlin

## Checklist

- JIRA link
- Check target branch

----


server/functional-test/src/test/java/org/zanata/feature/editor/EditorFilterMessagesTest.kt
```diff
@@ -0,0 +1,85 @@
...
+    @Test(timeout = MAX_SHORT_TEST_DURATION.toLong())
+    fun editorFilterIsBookmarkable() {
+        val urlForEditor = String.format(BasicWorkFlow.EDITOR_TEMPLATE, "about-fedora",
+                "master", "fr", document)
+        val urlWithFilterCondition = "$urlForEditor;search:hello%20w;resid:res1"
+        val editorPage = BasicWorkFlow().goToPage(urlWithFilterCondition,
+                EditorPage::class.java)
+
+        assertThat(editorPage.messageSources).containsExactly("hello world")
+        assertThat(editorPage.filterQuery.trim { it <= ' ' })
```
seanf: trim what?

server/functional-test/src/test/java/org/zanata/feature/editor/TranslateTextTest.kt
```diff
@@ -0,0 +1,122 @@
...
+    private val testFileGenerator = TestFileGenerator()
+
+    @Before
+    fun before() {
+        zanataRestCaller = ZanataRestCaller()
+        BasicWorkFlow().goToHome().deleteCookiesAndRefresh()
+        assumeFalse(
+                "",
+                File(CleanDocumentStorageRule.getDocumentStoragePath() +
+                        "${File.separator}documents${File.separator}").exists())
```
seanf: I know this is probably pre-existing logic, but wouldn't it be better just to delete the the directory, or simply fail the test? 

If this condition fails, I think the test will be skipped with little or no warning. That doesn't seem right.

djansen-redhat: Indeed, it's unnecessary - the rule throws if a delete fails.

server/functional-test/src/test/java/org/zanata/feature/editor/EditorFilterMessagesTest.kt
djansen-redhat: It actually trims the filter query, which adds a space on the end - but I've changed it to test against the _actual_ string. Good catch.

seanf: I'm not sure I really caught anything. I was trying to work out what exactly it was doing. Was it trimming any characters "less than or equal to" SPACE, ie control characters?

djansen-redhat:
Part of https://github.com/zanata/zanata-platform/pull/946

PR #946, opened 2018-07-09
Upgrade functional tests to JUnit5 and Kotlin

djansen-redhat:
JIRA issue URL: n/a

Use JUnit 5 (Jupiter) and Kotlin. Converts Rules to Extensions and Suites to Tags.

## Checklist

- JIRA link
- Check target branch

----


seanf:
@djansen-redhat I noticed that RetryRule.kt still exists, but it doesn't seem to be used. I don't think it's the sort of rule that works with JUnit 5, so it should probably be deleted. I think we only had it on TranslationHistoryTest before, and if I'm reading it correctly, `RetryRule(0)` is actually configured not to retry:

https://github.com/zanata/zanata-platform/blob/4d2fd34010c1e6762a60b2217bb57a1505b70a38/server/functional-test/src/test/java/org/zanata/feature/editor/TranslationHistoryTest.java#L43

But it made me think of something else: with JUnit 5 we lose an important Surefire feature (clunky though it is): https://maven.apache.org/surefire/maven-surefire-plugin/examples/rerun-failing-tests.html

Perhaps we can work around it by applying one of these solutions to all functional tests (specifically Selenium UI tests), preferably without putting the same annotation in every test file:

- https://github.com/junit-team/junit5/issues/1558#issuecomment-414694057
- https://github.com/artsok/rerunner-jupiter/
- https://stackoverflow.com/questions/46181026/junit5-how-to-repeat-failed-test

seanf:
@djansen-redhat I've started on some TestExecutionListeners in a branch: https://github.com/seanf/zanata-platform/tree/junit5-listeners

It's going to be tricky to get the test class names (and methods), to access their annotations, which would be helpful for `ScreenshotEnabledTestRunListener` and probably essential for `TestTracing`. 

But we might be able to do something with `org.junit.platform.launcher.TestIdentifier#getSource` if we can assume that all functional tests have a `MethodSource`. They probably will, unless we start using some sort of dynamic test generation. But flaky-retries might complicate that.

djansen-redhat:
- [ ] TimeoutException
- [ ] Async zanata rest caller functions should return the status, or at least have something to verify

PR #1004, opened 2018-07-29
fix(ZNTA-2721): batch requests to server

djansen-redhat:
JIRA issue URL: https://zanata.atlassian.net/browse/ZNTA-2721

Pushing 30k strings through MT proved to be excessive force. Batch to 1000.

## Checklist

- JIRA link
- Check target branch

----

### DIFF
--- a/server/services/src/main/java/org/zanata/service/impl/MachineTranslationServiceImpl.java
+++ b/server/services/src/main/java/org/zanata/service/impl/MachineTranslationServiceImpl.java
@@ -86,6 +86,7 @@ import static java.util.stream.Collectors.toList;
 public class MachineTranslationServiceImpl implements
         MachineTranslationService {
     private static final int BATCH_SIZE = 100;
+    private static final int REQUEST_BATCH_SIZE = 1000;
     private static final Logger log =
             LoggerFactory.getLogger(MachineTranslationServiceImpl.class);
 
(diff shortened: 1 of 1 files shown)


server/services/src/main/java/org/zanata/service/impl/MachineTranslationServiceImpl.java
```diff
@@ -86,6 +86,7 @@
 public class MachineTranslationServiceImpl implements
         MachineTranslationService {
     private static final int BATCH_SIZE = 100;
```
seanf: I guess this is SAVE_BATCH_SIZE now?

djansen-redhat: Done

server/services/src/main/java/org/zanata/service/impl/MachineTranslationServiceImpl.java
```diff
@@ -271,17 +272,29 @@ private MTDocument getTranslationFromMT(MTDocument request,
...
+            MTDocument mtDocument = textFlowsToMTDoc
+                    .fromTextFlows(projectSlug, versionSlug,
+                            doc.getDocId(), doc.getSourceLocaleId(),
+                            next,
+                            TextFlowsToMTDoc::extractPluralIfPresent,
+                            backendId);
+            MTDocument result = getTranslationFromMT(mtDocument,
+                    targetLocale.getLocaleId());
+            saveTranslationsInBatches(next, result, targetLocale, saveState);
+            backendIdConfirmation = result.getBackendId();
```
seanf: // TODO we only return the backendId from the final batch

I think the problem will go away with https://zanata.atlassian.net/browse/ZNTA-2732

djansen-redhat:
___
*[server/services/src/main/java/org/zanata/service/impl/MachineTranslationServiceImpl.java, line 291 at r3](https://reviewable.io/reviews/zanata/zanata-platform/1004#-LIZoF1Qhcy2EdzEcxvk-r3-291:-LItQLOa4bw5azV-8gox:b-bibs94) ([raw file](https://github.com/zanata/zanata-platform/blob/7349e5e3c7352f610cc2f08901ca9698e8da17e3/server/services/src/main/java/org/zanata/service/impl/MachineTranslationServiceImpl.java#L291)):*

    // TODO we only return the backendId from the final batch

I think the problem will go away with https://zanata.atlassian.net/browse/ZNTA-2732
</blockquote></details>

Good idea, I'll add the comment

PR #1009, opened 2018-07-31
fix(ZNTA-2727): TM Merge modal version filter check, remove componentWillReceiveProps

efloden:
JIRA issue URL: https://zanata.atlassian.net/browse/ZNTA-2727

Issue appears to be caused by the lack of null checking in a deprecated React lifestyle method componentWillReceiveProps.
Refactored the unsafe operation and the TMMergeModal to use the componentDidUpdate lifestyle method instead.

## QA
Issue affected the TM and MT merge modals when there exists a project without any versions.
Shouldn't experience any console errors when opening and running the TM and MT merge modals now.

## Checklist

- JIRA link
- Check target branch

----

### DIFF
--- a/server/zanata-frontend/src/app/containers/ProjectVersion/TMMergeModal.js
+++ b/server/zanata-frontend/src/app/containers/ProjectVersion/TMMergeModal.js
@@ -3,7 +3,7 @@ import React from 'react'
 import { Component } from 'react'
 import * as PropTypes from 'prop-types'
 import {connect} from 'react-redux'
-import { differenceWith, isEqual, throttle } from 'lodash'
+import { differenceWith, isEqual, isEmpty, throttle } from 'lodash'
 import {arrayMove} from 'react-sortable-hoc'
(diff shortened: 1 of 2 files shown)


spathare21:
@efloden  have to ignore codecov ?

efloden:
@spathare21 yes, ignore codecov for this one please (0.01%)

spathare21:
@efloden also if there is no document in the version TM and MT modal should have to open or not ?

efloden:
@spathare21 It looks like we do allow opening the modals when there are no documents for a version, though we probably shouldn't given both of the processes would not work without any target documents to merge with. Should we make another issue for this @aeng @seanf ?

seanf:
I don't think we **need** to do that check beforehand, although it would be nice to have. As long as the processes don't fail outright when there is no work to do. If they do, that's the more important problem.

aeng:
@efloden Lets make sure the MT merge does not fail over empty document version. We disable the option later for empty doc version.

PR #1013, opened 2018-08-02
fix(ZNTA-2730): clear hibernate cache before and after batch

djansen-redhat:
JIRA issue URL: https://zanata.atlassian.net/browse/ZNTA-2730

Use EntityManager:clear after processing a batch of textflowtargets in a transaction, to prevent progressively increasing memory usage.

## Checklist

- JIRA link
- Check target branch

----

### DIFF
--- a/server/services/src/main/java/org/zanata/service/impl/MachineTranslationServiceImpl.java
+++ b/server/services/src/main/java/org/zanata/service/impl/MachineTranslationServiceImpl.java
@@ -21,6 +21,8 @@
 package org.zanata.service.impl;
 
 import java.net.URI;
+import java.time.Duration;
+import java.time.Instant;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


djansen-redhat:
Uses the batch change from ZNTA-2721 (#1004) because large documents were a pain to work with.
This change decreases the allocated entities from 3000+ to 500 and processing time from about 10s  increasing to roughly 1.5s consistent.

Fails at the indicated line for attribution, with a session error. It may be a good idea, if possible, to move attribution to immediately after the batch, as it may run into a similar problem.

seanf:
@djansen-redhat Oh, you mean https://github.com/KengoTODA/findbugs-slf4j ? I wondered why you got rid of all the "[PERF]" prepending. SLF4J_FORMAT_SHOULD_BE_CONST was complaining, I assume.

djansen-redhat:
@seanf that, and after I removed it there was a missing parameter/list value in one of the calls so my approach was risky anyhow.

djansen-redhat:
@seanf #1004 was merged, and release merged into this. I noticed I added some whitespace to TranslationServiceImpl - should I remove it?

PR #1016, opened 2018-08-06
fix(ZNTA-2710) Increase MT progress granularity

djansen-redhat:
JIRA issue URL: https://zanata.atlassian.net/browse/ZNTA-2710

Use the overall TF count and batches to determine progress.

## Checklist

- JIRA link
- Check target branch

----

### DIFF
--- a/server/services/src/main/java/org/zanata/async/handle/MachineTranslationPrefillTaskHandle.java
+++ b/server/services/src/main/java/org/zanata/async/handle/MachineTranslationPrefillTaskHandle.java
@@ -25,8 +25,6 @@ import org.zanata.async.AsyncTaskHandle;
 import org.zanata.async.AsyncTaskKey;
 import org.zanata.async.UserTriggeredTaskHandle;
 import com.google.common.base.MoreObjects;
-import java.math.BigDecimal;
-import java.math.RoundingMode;
 
 /**
  * @author Patrick Huang
@@ -37,7 +35,7 @@ public class MachineTranslationPrefillTaskHandle extends AsyncTaskHandle<Void>
     private static final long serialVersionUID = 1704257523049053603L;
     private String triggeredBy;
     private String targetVersion;
-    private double percentage;
+    private long totalUnits;
     private long actualProgress;
 
     public MachineTranslationPrefillTaskHandle(AsyncTaskKey key) {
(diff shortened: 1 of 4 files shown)


server/services/src/main/java/org/zanata/service/impl/MachineTranslationServiceImpl.java
```diff
@@ -209,7 +210,10 @@ private MTDocument getTranslationFromMT(MTDocument request,
             log.info("No documents in {}", version.userFriendlyToString());
             return AsyncTaskResult.completed();
         }
-        taskHandle.setMaxProgress(documents.size());
+        // Set taskHandle to count the textflows of all documents
+        taskHandle.setUnits(
+                documents.values().stream()
+                        .mapToLong(d -> d.getTextFlows().size()).sum());
```
seanf: This could be very expensive (N+1 fetches) unless we write a custom query which returns all the project's text flows in one go (or better still, just the **number** of non-obsolete text flows in non-obsolete documents).

server/services/src/main/java/org/zanata/service/impl/MachineTranslationServiceImpl.java
```diff
@@ -263,15 +266,21 @@ private MTDocument getTranslationFromMT(MTDocument request,
...
+                                                         String backendId,
+                                                         AsyncTaskHandle taskHandle) {
         DocumentId documentId = new DocumentId(doc.getId(),
                 doc.getDocId());
         entityManager.clear();
         List<HTextFlow> textFlowsToTranslate =
                 getTextFlowsByDocumentIdWithConstraints(targetLocale,
                         documentId, overwriteFuzzy);
+        // Increase progress for non-translated items
+        taskHandle.increaseProgress(doc.getTextFlows().size() - textFlowsToTranslate.size());
```
seanf: So this, um, is the number of text flows we skip because they don't need to be translated? Maybe we should introduce a local variable `textFlowsToSkip` to make it clearer.

server/services/src/main/java/org/zanata/service/impl/MachineTranslationServiceImpl.java
```diff
@@ -301,6 +310,7 @@ private MTDocument getTranslationFromMT(MTDocument request,
             // TODO we only return the backendId from the final batch
             backendIdConfirmation = result.getBackendId();
             startBatch = batchEnd;
+            taskHandle.increaseProgress(next.size());
```
seanf: What's "next" again? We should give it a more descriptive name. textFlowsInBatch or something?

server/services/src/main/java/org/zanata/async/handle/MachineTranslationPrefillTaskHandle.java
```diff
@@ -38,6 +36,8 @@
     private static final long serialVersionUID = 1704257523049053603L;
     private String triggeredBy;
     private String targetVersion;
+    private double percentage;
```
seanf: We might be better off without introducing `percentage`.

djansen-redhat:
___
*[server/services/src/main/java/org/zanata/async/handle/MachineTranslationPrefillTaskHandle.java, line 38 at r2](https://reviewable.io/reviews/zanata/zanata-platform/1016#-LJMjyng-rfgoiKu6nEq:-LJqZ4zc-8xCUVqP6HL7:bjsjs2d) ([raw file](https://github.com/zanata/zanata-platform/blob/56c84af8edb9faf30986ff8128e4ae76fbbd8d7d/server/services/src/main/java/org/zanata/async/handle/MachineTranslationPrefillTaskHandle.java#L38)):*

How do these two values relate to currentProgress and maxProgress (in AsyncTaskHandle)?
</blockquote></details>

increaseProgress sets the handle currentProgress, from actualProgress (maybe unitsDone?) as a percentage of totalUnits.

djansen-redhat:
@seanf maybe too feature-y for release?

PR #1030, opened 2018-08-22
fix(ZNTA-2633): include JAXB jars

seanf:
JIRA issue URL: https://zanata.atlassian.net/browse/ZNTA-2633

Java 9 and 10 don't include JAXB on the classpath by default, and it's not included in Java 11 at all. So we need to add it as a dependency instead.

## Checklist

- JIRA link
- Check target branch

----

### DIFF
--- a/client/zanata-cli/pom.xml
+++ b/client/zanata-cli/pom.xml
@@ -56,6 +56,37 @@
   </build>
 
   <dependencies>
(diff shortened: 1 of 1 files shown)


djansen-redhat:
@seanf just to be clear, _building_ with javac 9/10 is a different issue, yes?

djansen-redhat:
Tested 8~11

PR #1032, opened 2018-08-23
Increase JS heap

seanf:

Apparently the default for Node is 1.76GB on 64 bit machines: https://stackoverflow.com/questions/47381275/why-does-node-js-have-an-1-76-gb-memory-limit?rq=1

This change increases it to 4GB when built via Jenkins.

## Checklist

- JIRA link
- Check target branch

----


djansen-redhat:
@seanf NO_QA, rebuild maybe 3-5 times and go?

seanf:
@djansen-redhat That's five builds in a row where the frontend module has finished without "JavaScript heap out of memory".

